PR #12, opened 2015-08-06
Correcting link to rockettheme.com

symac:
was missing a /


flaviocopes:
Thanks for pointing this out, made separate commit since there was a conflict with a commit I made in the meantime :)

Issue #48, opened 2015-08-11
Page-Edit: Could not go back from 'Expert' to 'Normal'


K232:
Hi all,
first: Thanks a lot for this great plugin!!

On the pages-menu I click on an existing page. I could click on all the header tabs. I also clicked on 'Expert' at the right side. The tabs are replaced by 'Content' and 'Options'. But nothing happens when I click on 'Options'. Same when I click on 'Normal' so I could not go back to the Normal-View. I also left the page and went back to the menu, clicked on another page, but there is the same result: It shows 'Content', 'Options', 'Normal' and 'Expert' but nothing happens when I click on one of them. Additionally I've noticed that the mouse pointer does not change when I hover one of these buttons, but it changes when I just move the mouse from 'Options' to 'Normal' on the tab row. It's changing to a 'drag-and-drop' mouse pointer, and I could hold the left button and drag one of my other pages appearing somehow.... Confusing to explain. If it does not happen on your installation please send me short info so I try to create a video of it.

Another thing: It would also be nice if the 'Please report it on Github' would contain a hyperlink to the repository.

Many thanks!!
Kai

flaviocopes:
Hi Kai, first question which browser are you using? 
Does the same happen with Chrome for example? If Yes, what's your setup? E.g. MAMP or anything else?

K232:
Hi Flavio,
I'm using Firefox 34 (oh, need to update...). I've tried with IE 10, and there the problem does not appear .... Additionally I've seen that the problem only exists when I'm working on new / existing pages within a certain folder. It's my first folder (01) and it contains a german umlaut ä . It has around 300 pages already. I also have another folder with an umlaut, there the problem does not appear....?? I've created a screen video at http://ge.tt/5VdVq0M2/v/0?c 

I've now upgraded to latest Firefox 39.0.3 on another machine. Same problem exists. But an easy workaround for me: Open an existing page which works fine and switch back to 'Normal' there. I think I don't need the expert mode.

flaviocopes:
Ok I tested the umlaut, I think we isolated the issue.

So Windows environment, Firefox fails, IE works fine.

I tested on Mac, works in both Chrome and Firefox. But, I cannot reach the page in the frontend if the folder/slug contains an umlaut.
Can you please check if you see the page in the frontend of the site? Or you get a 404 when trying to open it?

Another question: if you avoid putting the umlaut in the folder name, and instead change the page title / slug to contain it? e.g. 

[...]

PR #70, opened 2015-08-14
Fix for issue 48

K232:
If the list of posts is larger than 200 there might be technical problems when trying to edit a page, as seen in issue #48 . Therefore ordering is only available if the amount of post is less than 200.


K232:
#48

PR #77, opened 2015-08-18
Added German Language

kevinkub:
I have added all the german translations to the current version. However I have not been able to test some strings. I am willing to fix this in the next few days, when I have been able to check out all the texts in my installation and not only in editor.


flaviocopes:
Thanks @kevinkub, much appreciated! :+1: 
Will be included in the next admin plugin release!

Issue #146, opened 2015-09-07
Feature request: Allow plugins to hook into admin panel


Sommerregen:
At the moment the admin plugin does not allow adding custom (admin) routes, pages and menus injected and handled by plugins. It would be good to have certain hooks a plugin could be used i.e. `onAdminInitialized`, `onAdminPageInitialized`, `onAdminMenu` etc.

As far as I can see listing additional entries is possible via defining a [`template/partials/nav-pro.html.twig`](https://github.com/getgrav/grav-plugin-admin/blob/develop/themes/grav/templates/partials/nav.html.twig#L54), but I do not want to provide one, since this seems to be intended for the pro version.

Do I have other possibilities to show a menu entry in the Grav Admin panel and provide a custom page?

flaviocopes:
Was working right on this yesterday. 

I made an example plugin that plugs in the admin to provide its own view. It's an experimental thing for now. A couple of changes are needed in Admin: allow pages added by plugins, and a way to inject nav menu items.

We could have a `onAdminTemplateNav` (name hints welcome) event, to which we can provide the Admin plugin a label (already translated) and a route.

I'll post a PR today as a possible implementation

Sommerregen:
@flaviocopes Glad to hear! :-D I will test it as soon you and probably @rhukster provide a way for me to hook my own plugin into this. Thanks and keep me updated!

flaviocopes:
PR https://github.com/getgrav/grav-plugin-admin/pull/148 adds this possibility. Waiting for @rhukster's opinion if it's a good way to do it. Example plugin using this method: https://github.com/flaviocopes/grav-plugin-data

rhukster:
done

PR #148, opened 2015-09-07
Feature/allow plugins hook in admin

flaviocopes:
Adds 

1) the possibility for a plugin to provide pages to be used by the admin plugin

2) the ability to hook in the nav menu with the onAdminTemplateNavPluginHook event.

Example usage:

```
    public function onAdminTemplateNavPluginHook()
    {
        $this->grav['twig']->plugins_hooked_nav['PLUGIN_DATA.DATA'] = ['route' => 'data', 'icon' => 'fa-file-text'];
    }
```


flaviocopes:
Refs #146

PR #149, opened 2015-09-07
Adds a ignore_page_folders array to the admin config. Hides corresponding page folders in the pages twig template.

flaviocopes:
Pages count in the sidebar still includes hidden folders pages, might need to modify the pages object at a higher level to change that. 

Here's the simplest way to do it.

### DIFF
--- a/admin.yaml
+++ b/admin.yaml
@@ -16,3 +16,4 @@ popularity:
     daily: 30
     monthly: 12
     visitors: 20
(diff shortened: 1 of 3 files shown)


flaviocopes:
Refs https://github.com/getgrav/grav/issues/269

flaviocopes:
Better move in the Grav core

PR #152, opened 2015-09-07
Fix missing translation in Array field

Sommerregen:
This PR fixes a missing translation in the array field form, when this field is initialized with existing data.


flaviocopes:
Thanks!

Issue #140, opened 2015-09-04
Metadata og is broken to saving in Admin plugin.


Unrealno1:
Page metadata:
```
og:
    title: The Rock
    type: video.movie
    url: http://www.imdb.com/title/tt0117500/
    image: http://ia.media-imdb.com/images/rock.jpg
```

After, this saved in admin plugin.
```
og: 'The Rock, video.movie, http://www.imdb.com/title/tt0117500/, http://ia.media-imdb.com/images/rock.jpg'
```

flaviocopes:
I confirm this problem. Nested metadata is not correctly recognized / parsed by the admin plugin.

rhukster:
This will require a pretty sophisticated field to properly support.  Currently it uses an array, but would need to be almost like a dynamic tree.

rhukster:
i have an idea for a temporary workaround, that would at least not break the metadata

rhukster:
Flavio has promised to take a look at this one :)

flaviocopes:
WIP :hourglass:

flaviocopes:
https://github.com/getgrav/grav-plugin-admin/pull/155 is in, fixes the save problem and lets you edit the nested fields. You cannot add new nested metadata blocks for now, just edit existing nested metadata.

PR #155, opened 2015-09-08
Feature/fix nested metadata

flaviocopes:


flaviocopes:
screen shot 2015-09-08 at 11 37 05

flaviocopes:
Need to better position the `+` and `-` buttons at the bottom of the field.

There's not yet a way to add a nested field, maybe we can add a "fat" `+`, like the one used to add a nested field item.

flaviocopes:
Fixes #140

PR #165, opened 2015-09-13
Add russian translate

joomline:
Added language localization admin panel for version 0.5.0


rhukster:
Thanks for your hard work translating this!

PR #166, opened 2015-09-13
Added options field into pages

namaless:
This is important when we used into plugin for select one page, but if not used this every time we update one page (where we not need to setup this field) the field is saved into the page.

With this we use at the same of normal select so we can use the empty key with default text.

### DIFF
--- a/themes/grav/templates/forms/fields/pages/pages.html.twig
+++ b/themes/grav/templates/forms/fields/pages/pages.html.twig
@@ -2,6 +2,12 @@
 
 {% macro options(field, pages, value, depth) %}
 
+    {% if field.options and depth == 0 %}
+        {% for key, value in field.options %}
+            <option value="{{ key | e('html_attr') }}">{{ value }}</option>
+        {% endfor %}
+    {% endif %}
+    
     {% if field.show_root and depth == 0 %}
         <option value="/">/ (Root)</option>
         {% set depth = depth +1 %}


rhukster:
If your populating your own pages, does it not make more sense simply to use a 'select' field and pass in the options?

namaless:
This I need example into the simple_form plugin, if you not use any field not saved into the page, but with the pages field type the first is always one page so this saved into the page header.

namaless:
I have another way but this I think is the same of select so is more simple to reminder.

namaless:
This is perfect for me: https://monosnap.com/file/gIkLMzrsoOHU45bvP7BgDFACBonbse.png

You have another idea no problem, but with the pages field now we can't have this. Some other plugin used the same way have in the future the same problem.

namaless:
The other way for the select is this:

```php
    /**
     * @return array
     */
    public static function getPagesList()
    {
        $pages = Grav::instance()['pages'];

        $pages_list = $pages->getList($pages->root());

        $list = [];

        foreach ($pages_list as $route => $title) {
            /** @var Page $page */
            $page = $pages->dispatch($route, false);

            if (true === $page->routable()) {
                $list[ $route ] = str_replace(' -', '-', str_replace('&nbsp; &nbsp; ', '- ', $title));
            }
        }

        return $list;
    }
```
Choice if this way is much better then edit the admin pages field.

Issue #193, opened 2015-09-18
Saving changes opens off-canvas menu (mobile)


Mezarith:
The button for saving changes calls the off-canvas menu, before refreshing. I've only experienced this on mobile.

Mezarith:
UPDATE: all buttons and links that go to other pages within the admin site open the menu.

Vivalldi:
@rhukster and @Mezarith I've got this.

Mezarith:
You might want to add the bug label to this.
:P

Vivalldi:
no need as it's closed. It's more for tracking when open to show importance.

Mezarith:
@Vivalldi Ah, makes sense. I though the labels were also used for archives.

PR #198, opened 2015-09-22
Fixes #117 & #193

### LINKED ISSUE #117: Design is broken by language
I translated the language file in Japanese.
Admin Design is broken because the character width is different in Japanese.
e.g. no, expert

pages-option

Vivalldi:
Prevents wrong display of toggle button in languages such as japanese by removing the 50% width set on the buttons. #117


Vivalldi:
Also fixes #193 and some clean up to mobile.js

Issue #147, opened 2015-09-07
Enhancement: Support fieldset and/or accordion form field


Sommerregen:
For plugins with a many options it would be good to structure them. Currently this is possible with the `section` form field only. Is it possible to provide a grouping fieldset? Further it would be nice if such form field could be opened and closed like the one for the plugin list.

rhukster:
If you create the field, i'de gladly consider adding it.  Just look at section, tab, tabs, columns, etc.  These are all examples of structured field types.

Sommerregen:
Ok, this PR may take a while, since I'm currently running out of time. But I will add as soon as I have some minutes left.

PR #210, opened 2015-10-05
Add fieldset form field

Sommerregen:
This PR addresses #147 and adds a fieldset form field in the admin panel. Options are:

```yaml
fieldset.options:
  type: fieldset
  title: Your title
  help: Help text
  info: Info text
  text: Text inside fieldset and before other fields
  icon: comments
  collapsed: true     # Initial state of fieldset (see collapsible option)
  collapsible: true    # Whether one can expand the fieldset or not 

  fields:
   ... # like section form field
```

The fieldset comes with CSS styles. After merging the `template.css` needs to be re-compiled (I don't know your settings, thus I leave this task for you) in order to correctly display this new form field :-).

Issue #147 can then be closed.

### DIFF
--- a/themes/grav/scss/template/_forms.scss
+++ b/themes/grav/scss/template/_forms.scss
@@ -472,3 +472,64 @@ form {
         z-index: 10005;
     }
 }
+
+// Fieldset
+#admin-main .admin-block h2 {
+    font-size: 1.25rem;
+    margin: 0 0 .5rem;
+
+    letter-spacing: normal;
+}
+.form-fieldset {
+    background-color: #f7f7f7;
+    border: 2px solid #e1e1e1;
+    margin: 1rem 2rem;
+}
+
+.form-fieldset--label {
+    background-color: #f3f3f3;
+
+    &:hover,
+    .form-fieldset input:checked + & {
+        background-color: #eee;
+    }
+
+    label {
+        display: table;
+        font-size: 1.25rem;
+        padding: .5rem 1rem;
+        width: 100%;
+    }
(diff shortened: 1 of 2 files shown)


rhukster:
Could you post a screenshot of its usage? easier than me getting it setup :)

Sommerregen:
Of course. Here is the example with the above settings:

example

And here a live preview:

demo

flaviocopes:
Would be nice if we could find a general way to make the fieldset disappear when the above toggle is disabled. Could simplify some settings here&there. Just throwing in my idea, not sure if it's useful or even if it has usability implications

Sommerregen:
@flaviocopes Can you explain a little bit more, which toggle you mean? Do you mean `collapsible`? In this the arrow on the right vanishes and it will always stays open. You can see it in the above screenshot (Summary fieldset).

flaviocopes:
For example, clicking 'No' on "Markdown" could entirely hide all the sections that are not relevant if that option is 'No'. It's not specific to this case, it's a general thing I had in mind.

Examples of places where this can be applicable:

screen shot 2015-10-06 at 18 59 27
screen shot 2015-10-06 at 18 59 19

Sommerregen:
@flaviocopes As far as I know and reverse engineered this is possible by now. Maybe I'm mistaken but, when programming the fieldset I noticed a `toggleable` option (https://github.com/getgrav/grav-plugin-admin/blob/develop/themes/grav/templates/forms/field.html.twig#L13-L24), which does quite what you want.

PR #235, opened 2015-10-26
Minor improvements

Sommerregen:
This PR improves resolving (plugin dependent) admin pages using the stream locator instead of a hardcoded file path (to make it multisite capable!). Further it changes the location where admin pages have to be placed in a plugin. Currently on pages of plugin are picked up, when they are in the `user/plugins/<plugin>/admin/pages` folder. With this PR the location of admin pages changes to e.g. `user/plugins/<plugin>/pages/admin` (note the change in the last two parts to conform to the admin page structure of the admin plugin).

Further I'd like to rename the hook `onAdminTemplateNavPluginHook` to `onAdminMenu`, which is shorter and does not look cumbersome. To be honest this is a matter of taste. If you don't like, I'll will delete it back.

All my changes are backward-compatible. Once you are sure, that nobody uses the deprecated functions/page structure, you can delete them.

### DIFF
--- a/admin.php
+++ b/admin.php
@@ -195,6 +195,16 @@ class AdminPlugin extends Plugin
             $plugins = Grav::instance()['config']->get('plugins', []);
 
             foreach($plugins as $plugin => $data) {
+                $path = $locator->findResource(
+                    "user://plugins/{$plugin}/pages/admin/{$self->template}.md");
+
+                if (file_exists($path)) {
+                    $page->init(new \SplFileInfo($path));
+                    $page->slug(basename($self->template));
+                    return $page;
+                }
+
+                // DEPRECATED: Will be removed!
                 $folder = GRAV_ROOT . "/user/plugins/" . $plugin . "/admin";
 
                 if (file_exists($folder)) {
@@ -245,6 +255,9 @@ class AdminPlugin extends Plugin
         $twig->twig_vars['admin'] = $this->admin;
 
         // Gather Plugin-hooked nav items
+        $this->grav->fireEvent('onAdminMenu');
+
+        // DEPRECATED
         $this->grav->fireEvent('onAdminTemplateNavPluginHook');
 
         switch ($this->template) {


rhukster:
I'll discuss this with @flaviocopes tomorrow.  I'm ok with the changes as long as they don't impact existing stuff, or if they do, they can be addressed without much effort.

flaviocopes:
The stream locator change is fine to merge
Event name change: can be merged, not sure it's worth the deprecation but I like the event name more than the one I used.

The plugin folder structure was added in this way to clearly separate when a plugin offers both frontend and admin capabilities.

For example the Comments plugin has frontend functionality, then under admin/ there's all (and just) the admin-related files. I see the admin plugin does things differently for pages, but then we also have templates to handle, if the plugin is adding templates in both the admin and the frontend it can be better to separate them clearly, instead of having (e.g.)

```
templates/
  partials/
  admin/
  admin/partials/
pages/
  admin/
```

we have

```
admin/
  pages/
  templates/
templates/
  partials/
```

Also, usually pages are only added for the admin-part, since in the frontend pages are created by the user, so I thought it was better to "hide" them under `admin`. 

That's the explanation of why it's like that, if you have more arguments on changing the structure it can be discussed.

Sommerregen:
@flaviocopes Thanks for explanation :+1: . I only have seen how the admin plugin structures the pages and templates and was wondering why it is handled differently for plugins. I understand your point, that the structure like

```php
templates/
  partials/
  admin/
  admin/partials/
pages/
  admin/
```

is more confusing, then the one implemented already. I can remove it, no problem at all! :smile:

Concerning deprecation, I thought it is the way do to: introduce something new, mark the old one as deprecated and then delete the deprecated stuff in the next version. But we/I can delete it, if you like, but then you have to update your plugins. The same goes for changing the event names. Don't know what @rhukster is thinking about and what's the better way...

BTW: The `Data manger` has a long and really ugly event name `onDataTypeExcludeFromDataManagerPluginHook`, too, which can be shortened to `onDataMangerExludeDataType`. I think (and guess this was the way how events were named in grav) it is good to stick to some rule like

```
on<pluginName><eventName>
```

rhukster:
Going to look at merging this after Admin plugin RC3 is released (soon).

flaviocopes:
Thanks! Added your locator changes to the existing structure

Sommerregen:
Mh, @flaviocopes oh I forgot! After our discussion I wanted to stick to your folder structure, since it is more clearer. Can you change this (line https://github.com/getgrav/grav-plugin-admin/pull/235/files#diff-2a524880b4d504a3da21007051e59146R199)

```php
$path = $locator->findResource("user://plugins/{$plugin}/admin/pages/{$self->template}.md");
```

and delete the other stuff below?

flaviocopes:
I pushed a PR containing your commit, already did that :+1: (should have appeared here!) https://github.com/getgrav/grav-plugin-admin/pull/245/files

Sommerregen:
@flaviocopes Ah thanks and sorry sorry for any inconvenience :relaxed:

flaviocopes:
np!

PR #253, opened 2015-11-04
Escape values in `array.html.twig` form fields

Sommerregen:
This PR addresses a minor issue with the `array` form fields. Up to now array values ar not allowed to contain double apostrophes, otherwise the field in the admin panel breaks and yields invalid HTML markup. This PR proposes to escape such values in order to display them correctly. As far as I tested escaping has no side-effects, meaning double apostrophes are posted unescaped back.


rhukster:
Thanks benny!

PR #261, opened 2015-11-07
Fix saving on Edge

810:


w00fz:
Thanks looks good.

For future reference: https://dev.windows.com/en-us/microsoft-edge/platform/status/formattribute

PR #287, opened 2015-11-20
Hide "Latest Pages" if there aren't any pages available

Sommerregen:
This PR addresses a minor issue installing the admin plugin without having any pages in the `user/pages` directory. At the moment logging in yields to an error in `dashboard.html.twig` apparently because `admin.latestPages` returns null, which isn't iterable. This PR check the value and suppresses this issue.


flaviocopes:
:+1:

PR #291, opened 2015-11-21
Update CHANGELOG.md

diomed:
hr is not hungarian, but croatian translation [hr = hrvatski - croatian, that is]

my bad - i see hungarian is added as well, but croatian was not mentioned at all. so...


flaviocopes:
Ah that was probably missed in the changelog!

PR #329, opened 2015-12-06
Feature/improve plugin settings

flaviocopes:
This PR makes more space on the Plugin options, and allows a tab-based plugin option organization, like the one we can have for Pages.

screen shot 2015-12-06 at 17 54 14

### DIFF
--- a/themes/grav/templates/partials/plugins-details.html.twig
+++ b/themes/grav/templates/partials/plugins-details.html.twig
@@ -12,7 +12,26 @@
 {% endif %}
 </div>
 
-<br />
+{% if plugin.form.fields.tabs %}
+    <br>
+{% else %}
+    <h1>
+        {{ plugin.name|e }}
+        {% if admin.isTeamGrav(plugin) %}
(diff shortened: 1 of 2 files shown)


flaviocopes:
@rhukster left here to evaluate

rhukster:
Is this an option that can be configured in blueprints, or is it this way for all? Where is the plugin details now?

flaviocopes:
The plugin details shown now are the same ones that appear in the drop-down in the plugins list.
But maybe your hint is right, it can be configurable in the blueprint for each plugin.

flaviocopes:
The logic of my change was: this is currently, when adding tab-based navigation for the plugin settings. The details take a lot of space, and the tabs don't make a lot of sense in the middle of the page.

screen shot 2015-12-06 at 18 58 27

flaviocopes:
Synced the PR to automatically drop the plugin details when the plugin has tab-based settings, good idea?

rhukster:
I think the details should be in a tab if this is the case.

flaviocopes:
Nevermind! Will adjust a bit the page CSS and leave the details on top

PR #373, opened 2015-12-30
Added missing data-grav-field=select to global attributes on pagemediaselect field template

aymerix:
I tried to use pagemediaselect field inside a Page form but data from this field were not retrivied by the AdminController. However, the themeselect field was working so I compared files and I got pagemediaselect working by adding this line :

```
data-grav-field="select"
```

inside the block "global_attributes" in the pagemediaselect.html.twig template.

I'm new to grav cms so I hope this is the right way to fix the issue I was facing.

### DIFF
--- a/themes/grav/templates/forms/fields/pagemediaselect/pagemediaselect.html.twig
+++ b/themes/grav/templates/forms/fields/pagemediaselect/pagemediaselect.html.twig
@@ -8,6 +8,7 @@
 
 {% block global_attributes %}
     data-grav-selectize="{{ (field.selectize is defined ? field.selectize : {})|json_encode()|e('html_attr') }}"
+    data-grav-field="select"
     {{ parent() }}
 {% endblock %}
 
@@ -37,5 +38,3 @@
         {% if admin.page.media.all is empty %}Add files through the page media, or by dropping them in the page folder{% endif %}
     {% endif %}
 {% endblock %}
-
-


flaviocopes:
Strange, it works for me. Just tried again. 
Can you paste the whole blueprint where you added that, just to check what might be wrong?

aymerix:
Of course ! Here's my blueprint file. I've tested with Google Chrome on Linux, don't know if browser changes anything.

```
title: Shoot
'@extends':
    type: default
    context: blueprints://pages

form:
    fields:
        tabs:
            type: tabs
            active: 1

            fields:
                content:
                    fields:

                        preview:
                          type: section
                          title: Shoot Preview
                          underline: true

                        header.preview:
                            label: Choose media
                            type: pagemediaselect
```

flaviocopes:
Works fine to me, not sure what could be wrong in your case. Are you using latest Grav and latest Admin plugin, right? 
Anyway if this change makes it work for you, going to see exactly why. I thought I could recreate the problem, not sure why not.

aymerix:
I'm using Admin v1.0.4 and Grav v1.0.6.

Do you know why themeselect use "data-grav-field=select"  in "global_attributes" block and not pagemediaselect ?

I'm using this blueprint from a custom theme by the way.

flaviocopes:
I included the PR, that line was not included as I migrated from the `select` field, you're right.

I improved the field code in https://github.com/getgrav/grav-plugin-admin/commit/f34a9d7638fd86d9be129b8b9fb37f85dd341ea8 to to avoid double inclusion of `data-grav-field` (which was now occurring), and to disable the field when no files are found

aymerix:
Great ! Thank you for your great work :)

flaviocopes:
Ping if you still have issues with it!

PR #382, opened 2016-01-05
Fixed a few missing translations on /admin/info page

martinsik:
... in [info.html.twig](https://github.com/getgrav/grav-plugin-admin/blob/develop/themes/grav/templates/info.html.twig)


flaviocopes:
Thanks for taking care of this, actually we're reworking this part in https://github.com/getgrav/grav-plugin-admin/pull/377, to be merged soon, and this twig file will be removed (and the translations fixed)

flaviocopes:
So I'm closing this PR

PR #405, opened 2016-01-22
Fixed problem with detecting ctrl key on Windows

solusipse:
This contribution fixes an error that is probably related to jquery's bug - event.ctrlKey triggers not only ctrl but also right ALT key. That might cause some troubles for users with keyboard layouts that requires pressing `right alt + s` for inserting regional characters. I could recreate this issue only on Windows machines.

There's little snippet for those who would like to test this on different platforms:
```
<script src="https://code.jquery.com/jquery-git2.min.js"></script>
<script>
$(window).on('keydown', function(event) {
    if (event.which == 17) {
        alert("ctrl");
    }
    if (event.which == 18) {
        alert("alt");
    }
});
</script>
```

### DIFF
--- a/themes/grav/js/admin-all.js
+++ b/themes/grav/js/admin-all.js
@@ -527,7 +527,7 @@ $(function () {
     if (saveTask.length) {
         $(window).on('keydown', function(event) {
             var key = String.fromCharCode(event.which).toLowerCase();
-            if ((event.ctrlKey || event.metaKey) && key == 's') {
+            if ((event.which == 17 || event.metaKey) && key == 's') {
                 event.preventDefault();
                 saveTask.click();
             }


solusipse:
I just got the response from jquery team. This error is specific for Polish Programmer keyboard layout (most popular among Polish layouts), so I don't know if you'll want to merge it.

flaviocopes:
Needs to be tested on non-polish windows keyboards layouts, don't have a windows machine at hand now, someone can confirm?

shadhar:
I have the same problem... Right Alt + S saves the page, but I can't write in Polish. I've encountered the same problem in various other products so far.. like Brackets... it might be a good idea to an option to disactivate this binding or create a different one.

And other than that grav i amazing! Thank you!

flaviocopes:
@hexplor do you have a change to thest the PR on a polish keyboard layout? @rhukster or should we disable this saving enhancement if we detect polish is enabled?

shadhar:
Well there are so many other languages that require the use of Right Alt + S for special characters. My shot is that it'd be a good idea to provide a config yaml file that allows any user to define their own shortcuts...

For the time being it'd awesome to know which part of admin-plugin's code defines the shortcuts,so I could change it myself.

I know that updating the plugin would override my changes, but I can just go there and change it again, right?

w00fz:
I'm going to take a look at this today. Although note that there is an ongoing rewrite of the admin js in the `es6_revamp` branch so this will likely get fixed there as it's close to finalization and merging.

I don't like the idea of a configuration for this, cmd+s, ctrl+s are universal shortcuts for saving so I'd rather get it universally working on polish keyboards as well.

The fact that the jQuery team knows about it and hasn't fixed it is not a good sign, might not be a simple fix, but I'll keep you updated on it as soon as I get a better picture.

For manual temporary fix, you can patch the file directly. Look in the changes tab of this pull request to see what file @solusipse changed and how.

// Djamil Legato // From iPhone //

shadhar:
Thanks for your response @w00fz  - very helpful! I've changed the shortcut and can keep on writing my article woo hoo!

w00fz:
Unfortunately your fix is not going to work @solusipse, by checking the `which` for the CTRL like you are doing, you are preventing the `s` key detection. The `event.which` is always only one value per keydown event emitted. 

This is why we need to use the `event.[meta|alt|ctrl|shift]Key`. Whatever key letter is pressed, can be checked against the modifier keys which are preserved when the event is emitted.

The behavior is indeed very weird in polish programmer layout, pressing the right ALT key is what triggers the left CTRL. This I believe is because on Windows, the right ALT key is actually ALTGr, which is a combo of CTRL + ALT.

I have created a fiddle here https://jsfiddle.net/w00fz/34jc1vn9/6/ with a potential resolution that seems to be working fine for me, if you guys could give it a try and confirm, that would be very useful.

Thanks.

shadhar:
Works like a charm. I also checked in admin-all.js file. I can save with Ctrl+s and write 'ś' with Right Alt+s. Thanks @w00fz !!

w00fz:
Closed this PR since it's been fixed in a different way.

Thank you all for helping resolving this one!

solusipse:
@w00fz Yep, haven't noticed that. Your fix works great, thanks!

Issue #407, opened 2016-01-24
Absolute theme_url, not working multisite config, missing js, css assets


hotovec:
Hey. Check this line:
https://github.com/getgrav/grav-plugin-admin/blob/develop/admin.php#L421

It is absolute path. In case you are trying configure multisite grav.
Assets like js, css etc are not loaded to page. 
![screenshot 2016-01-24 02 58 20](https://cloud.githubusercontent.com/assets/388276/12534001/5d2b64b4-c246-11e5-9ca8-f315c194f46c.png)

I'm using subdomain config from Grav docs.
My structure: 
![screenshot 2016-01-24 02 54 16](https://cloud.githubusercontent.com/assets/388276/12533998/461cb2dc-c246-11e5-83ad-8574cf97c779.png)

rhukster:
Should be sorted with my last fix.

PR #408, opened 2016-01-24
Dynamic Route for theme

Vivalldi:
Deals with #407


Vivalldi:
I don't have a multi-site setup so haven't tested there

rhukster:
https://github.com/getgrav/grav-plugin-admin/commit/086ed34bc0e0b35c7edfb69f888813625d6c65ed

I actually have a safer fix.

PR #431, opened 2016-02-27
Updated taskSave to ensure route is prefixed with a forward slash

gRoberts84:
Came across this issue when attempting to add a modular page to my parent page.

`$route` would return the correct slug for the parent page, however without a leading forward slash, so it didn't match any of the routes within `$this->routes` resulting in the `$parent` object being `NULL` and throwing an exception.

When watching the `$pages->dispatch` method being called, all other routes were prefixed with a forward slash, so thought it would be best to always ensure the route being checked had a leading forward slash.


rhukster:
Good find! Thanks.

PR #438, opened 2016-03-03
Added list field option to set add item label

### LINKED ISSUE #409: Add blueprint list (type) button label override and lang translation
## Features added
It would be nice to have an override field for the 'Add Item' button in the blueprint form field type 'list'.

* https://github.com/getgrav/grav-plugin-admin/blob/develop/themes/grav/templates/forms/fields/list/list.html.twig#L67
* https://github.com/getgrav/grav-plugin-admin/blob/develop/themes/grav/templates/forms/fields/list/list.html.twig#L107

The  'Add item' default should also be added to be translations in the lang file.
Thanks
Shawn

gRoberts84:
Updated List field to allow users to override the `Add item` button label as requested by a user on Gitter and also to resolve #409


flaviocopes:
Thanks!

PR #458, opened 2016-03-23
Only display the translated logout message

tyrion273:


flaviocopes:
:+1:

tyrion273:
Are you OK with it? Because at first it worked but somehow the message is not displayed at all.

tyrion273:
I think something is buggy because after inserting the line again the message is displayed again but only in English although my backend language is German.

flaviocopes:
No issues for me with latest develop branches of grav+admin

PR #486, opened 2016-04-08
Different branches of Chinese characters

dommyet:
Use "zh" itself for Chinese is absolutely OK however there are different types of Chinese character. I believe there are some simplified Chinese users having difficulties reading traditional Chinese and vice versa.

The original zh.yaml is written in traditional Chinese, it should be either zh-hk.yaml or zh-tw.yaml, and there should be a simplified Chinese version named zh-cn.yaml. 

Language "zh" contains several branches, also there are two different standards.
The first one is the ISO 639-1, language zh has two branches which are "zh-Hans" for simplified Chinese and "zh-Hant" for traditional. I tried to put one of them into system.yaml however it can't be detected correctly.
The second standard (refer to https://msdn.microsoft.com/en-us/library/ms533052(v=vs.85).aspx) could work fine with Grav. They are "zh-cn" for simplified, "zh-hk" for traditional Hong Kong version, "zh-tw" for traditional Taiwan version and "zh-sg" for Singaporean version.


flaviocopes:
Translations are managed through Crowdin now https://crowdin.com/project/grav-admin, and merged in the plugin when doing a release, so any change will need to be made there.

I suggest to also discuss this matter on the Translation chat: https://gitter.im/getgrav/translations

PR #505, opened 2016-04-20
fixes #412

### LINKED ISSUE #412: User can't change his profile with certain ACL
Hi,

I've tried tu use ACL for a particular user who's simply going to add pages and edit them.
I gave this user certain permissions like this:
```
admin:
    login: true
    pages: true
```

It turns out, this user can't change his profile info, permission denied...
I'm guessing this is not the expected behaviour ?

Vivalldi:
Just saw this pop up in my stream and seemed like it should be easy enough...
Let me know if there are any issues

### DIFF
--- a/pages/admin/users.md
+++ b/pages/admin/users.md
@@ -2,6 +2,7 @@
 title: Users
 
 access:
+    admin.login: true
     admin.users: true
     admin.super: true
 ---


flaviocopes:
`users.md` is actually also used in the User Manager in Admin Pro, that's why it has those permissions, need to find another way to fix this. Maybe have a single `user.md` page for own profile.

flaviocopes:
Otherwise all users with access to the Admin would now see the user manager.

Vivalldi:
Ohhh that makes sense. Forgot about the pro features.

flaviocopes:
Closing this now, but marked on my list

Issue #394, opened 2016-01-15
Copying page in Admin makes resulting page have same slug


paulhibbitts:
When copying a page the slug of the new page does not seem to get changed, and is the exact same slug as the original page which then causes various issues.

I've attached a screenshot, look at the new copy folder name field vs. slug field.

![edit grav cms for educators flipping your lms with an open and collaborative web platform designing and building open source learning environments 2016-01-14 17-21-06](https://cloud.githubusercontent.com/assets/1812771/12342884/a8c36d28-bae3-11e5-96cd-f12f3b310089.png)

To repeat issue:
1. Open existing page
2. Click "copy" button
3. Go back to page list
4. Open newly created page
5. View slug on "Advanced" pane

Thanks very much,
Paul

flaviocopes:
What you suggest, adding `*-copy` to the slug?

paulhibbitts:
Hi Flavio,

I think we want to make it align with the current naming scheme logic, which seems to be using a " - #" format where # is either 2 or adding 1 to the existing value.

So, let's say there is a page called "New Page" and we make a copy. Here is what I would expect:

Copied page
- title: "New Page 2" (currently it is not, just the same title as orig - I suggest we change it)
- folder : "new-page-2" (currently as-is)
- slug: "new-page-2" (currently it is not, just the same slug orig. This is not an issue if no slug is defined on original page as new page slug field is then left blank)

What do you think?

Thanks very much,
Pau

flaviocopes:
:+1: sounds good!

flaviocopes:
Created a PR to solve this. Works at the basic level, still need to test with multilanguage and edge cases.

flaviocopes:
Doing final tests to merge this

flaviocopes:
Fixed by https://github.com/getgrav/grav-plugin-admin/pull/537

flaviocopes:
@paulhibbitts give it a try when you have time, just to make sure there are no issues I didn't find

paulhibbitts:
This is really awesome @flaviocopes! I've been able to do some testing and looks like a winner 👍

This is a perfect way to blow the mind of a WordPress person... 'you can even have two or more blogs in one Grav site, look it's so easy just a simple page copy and you're done.' 😉

paulhibbitts:
So everything is working really nicely, but I think there will  be less possible user confusion if the new title of copied pages is also updated.

[...]

Issue #429, opened 2016-02-23
Number prefix is not updated on page copy


jjui:
When copying a page with number prefix, the same number is applied to the new page-folder but adds new number on end of folder. 

Worth a look!

flaviocopes:
Linked to https://github.com/getgrav/grav-plugin-admin/issues/394

flaviocopes:
Linked to https://github.com/getgrav/grav-plugin-admin/issues/394

flaviocopes:
@rhukster thoughts on what's the desired behavior on this?

flaviocopes:
On it

flaviocopes:
Fixed by merged PR #537

jjui:
Nice 👊

flaviocopes:
Give it a try when you have time, just to make sure there are no issues I didn't find

PR #537, opened 2016-05-02
Re #394 update slug, route and raw route with appended number when doing a copy of the page

flaviocopes:
Does not change the page title, just the page route property if set in
the header, to avoid two pages with the same slug / route / raw route.


flaviocopes:
Work in progress. Works at the basic level, still need to test with multilanguage and edge cases.

flaviocopes:
Should now also solve https://github.com/getgrav/grav-plugin-admin/issues/429

rhukster:
I think there's a minor issue.

I created a new page with folder `new-page`, and set a custom slug to `yah-baby`.  Then i copied the page.  

The resulting page was created in a folder called `yah-baby-3` and the slug was set to `yah-baby-2`.  Expected results would be `new-page-2`.

flaviocopes:
Replicated

flaviocopes:
Updated with fix. Covers my tests up to now.

Issue #567, opened 2016-05-06
Username not substituted in error message in Forgot Password


Quy:
`User with username %s does not exist`

I got this error message with the `admin` username which is a valid account.

PR #568, opened 2016-05-06
Fix forgot password / reset password. Fix Twig escape in email template

flaviocopes:


flaviocopes:
Refs #567

rhukster:
I've fixed this slightly differently in combination with another fix.

PR #596, opened 2016-05-17
Fix file field

flaviocopes:
Used data[fieldname] but there's a bug with square brackets and $_FILES

https://bugs.php.net/bug.php?id=65763


rhukster:
This does fix the initial issue, but my PR takes into account a few other things that were broken. This one is redundant now, so closing.  Thanks though! :)

PR #615, opened 2016-05-23
Provide security options for single tabs

Sommerregen:
This PR allows to attach security options to single tabs (`tab`). Just like `section`.


mahagr:
Using security option like this will cause data to be lost if user doesn't have permissions to save all the data.

PR #638, opened 2016-06-02
Handle special case of media configuration

flaviocopes:
Since it's a list of dynamically generated entries, without parent, we
cannot merge in the post data with the defaults, or the defaults are
always saved, and we cannot remove existing fields


classes/admin.php
```diff
@@ -328,8 +328,19 @@ public function data($type, array $post = [])
...
             $blueprints = $this->blueprints("config/{$type}");
             $config = $this->grav['config'];
-            $obj = new Data\Data($config->get($type, []), $blueprints);
+
+            if ($type === 'media') {
+                if (count($post) === 0) { //reading
+                    $obj = new Data\Data($config->get($type, []), $blueprints);
+                } else { //saving
+                    $obj = new Data\Data([], $blueprints);
+                }
```
w00fz: This could probably just be 

```php
if ($type === 'media' && count($post)) {
    $obj = new Data\Data([], $blueprints); // saving
} else {
    $obj = new Data\Data($config->get($type, []), $blueprints);
}
```

rhukster: I agree, that would be equiv.

Issue #647, opened 2016-06-04
Add folder/modular missing text


JesseVent:
Description:
Looks like the "Add Folder" and "Add Modular" buttons inside the dropdown-menu aren't inheriting the CSS properly for view ports 59em and below.

Platforms attempted:
Windows 10 - Chrome
iOS 9.2 iPhone 6 Plus - Chrome
iOS 9.2 iPhone 6 Plus - Safari

Latest RC distribution

Steps taken: 
Logged in to admin
Cleared cache
Went to manage pages
Clicked on add page dropdown
Empty box appears (still clickable buttons though)

PR #648, opened 2016-06-04
Responsive button display error

JesseVent:
This fixes the add folder/modular missing text issue #647 for the add folder/add module text not displaying when viewing display ports 60 or under.
button font-size class was set to 0 for mobile.


rhukster:
Fixed slightly differently. cheers.

Issue #674, opened 2016-06-24
Feature suggestion: allow admin login with username OR email address


a1studmuffin:
Just a minor feature suggestion, it would be great if the Grav admin login screen allowed you to enter your admin username OR email address. Some browsers cache the admin email and password from the initial "create an account" screen, then when you advance to the login screen for the first time, the browser will auto-fill the form with the incorrect data. It made me scratch my head for a moment before figuring out what was happening. Cheers :)

a1studmuffin:
Wow that was quick! You guys rock, very impressed with the Grav developer community. Thanks :)

flaviocopes:
Thanks to @gsumpster

PR #685, opened 2016-07-04
Feature: Ability to Login with Email

gsumpster:
Implemented #674, added utility class to house `findUserbyEmail($email)`, created email regex check in `classes/admin.php`'s `authenticate($data, $post)` function.

### DIFF
--- a/admin.php
+++ b/admin.php
@@ -290,6 +290,9 @@ class AdminPlugin extends Plugin
             };
         }
 
+        // Load utility class
+        require_once __DIR__ . '/classes/utils.php';
+
         // We need popularity no matter what
         require_once __DIR__ . '/classes/popularity.php';
         $this->popularity = new Popularity();
--- a/classes/admin.php
+++ b/classes/admin.php
@@ -14,6 +14,7 @@ use Grav\Common\Themes;
 use Grav\Common\Uri;
 use Grav\Common\User\User;
 use Grav\Common\Utils;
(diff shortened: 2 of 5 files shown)


flaviocopes:
@rhukster you think we can merge this feat in 1.1 or we do it in a point release after it?

rhukster:
I think we can merge if all works as expected. Not tested myself.

Cheers!

--
Andy Miller
Sent with Airmail

admin.php
```diff
@@ -290,6 +290,9 @@ public function onPluginsInitialized()
             };
         }
 
+        // Load utility class
+        require_once __DIR__ . '/classes/utils.php';
```
flaviocopes: Use `use Grav\Common\Utils;` at the top, or there's a problem (try installing the Form plugin, for example, and you'll see)

flaviocopes: Sorry, misread - it's a new file. But it triggers the problem.. do we really need a separate file just for `findUserbyEmail`?

flaviocopes: Added my change in https://github.com/getgrav/grav-plugin-admin/pull/686/commits/b33fd299fe31648878e4b6c5d0b683ee7ed7d892, what you think?

flaviocopes:
Did a couple fixes in #686 and merged in https://github.com/getgrav/grav-plugin-admin/commit/64a88c916cb048896582ae7433293bb82728131f, thanks!

Issue #411, opened 2016-01-28
Allow permission groups to hide/show page editor tabs


nineteeneightyeight:
It would be quite useful if we could lock down certain tabs from certain admin user groups. 'Advanced' tabs for example should only be visible to advanced admin users.

flaviocopes:
We could add an access property to the tab and analyze it when rendering the tabs. 
All tabs need to be always present in the page I think to be saved correctly (need to check), but we could at least hide them.

mahagr:
Implemented in Admin 1.10.

PR #712, opened 2016-07-20
Hide tab when user is not authorized to access it

Sommerregen:
This PR references https://getgrav.org/forum#!/general:hiding-certain-tabs-from-no and #411 and hides the tab, when the user is not authorized to access it.


flaviocopes:
👍

PR #780, opened 2016-09-17
Language ES fixed

pixelead0:
The translation of 'language' to spanish is 'idioma'


rhukster:
Thanks for the fix, but translation changes need to be done via crowdin: http://translate.getgrav.org/project/grav-admin

Please register and make the edits there to ensure they get included in future releases.

PR #781, opened 2016-09-17
Default headers

stephencookdev:
This PR does the following:

 * Adds an "Frontmatter" section to the grav-admin plugin options
 * Adds an editor option to set some initial headers into a new page
 * Fixes a minor bug that stopped the "help" text from being readable

This will allow headers to be specified in plugin options for the admin plugin, and (unless overridden) will be present on all new page creations (through the admin plugin)

The specific use-case in mind is to add automatically the
```
access:
  site.login: true
  admin.login: true
```
headers required by the login plugin for pages to be log-in only

### DIFF
--- a/blueprints.yaml
+++ b/blueprints.yaml
@@ -321,4 +321,23 @@ form:
       label: Visitors history
       default: 20
 
+    Editor:
+      type: section
+      title: Editor
+      underline: true
 
+    editor.default_headers:
+      classes: frontmatter
+      type: editor
+      label: Default Headers
+      size: large
+      codemirror:
+        mode: 'yaml'
+        indentUnit: 4
+        autofocus: true
+        indentWithTabs: false
+        lineNumbers: true
+        styleActiveLine: true
+        gutters: ['CodeMirror-lint-markers']
+        lint: true
+      help: Default headers to add to new pages
--- a/blueprints.yaml
+++ b/blueprints.yaml
@@ -329,7 +329,7 @@ form:
     editor.default_headers:
       classes: frontmatter
       type: editor
-      label: Default Headers
+      label: Default Page Headers
       size: large
       codemirror:
         mode: 'yaml'
@@ -340,4 +340,4 @@ form:
         styleActiveLine: true
         gutters: ['CodeMirror-lint-markers']
         lint: true
-      help: Default headers to add to new pages
+      help: Default page headers to automatically add to any new page created
--- a/blueprints.yaml
+++ b/blueprints.yaml
@@ -321,12 +321,12 @@ form:
       label: Visitors history
       default: 20
 
-    Editor:
+    Frontmatter:
       type: section
-      title: Editor
+      title: Frontmatter
       underline: true
 
-    editor.default_headers:
+    frontmatter.default_headers:
       classes: frontmatter
       type: editor
       label: Default Page Headers
--- a/classes/admin.php
+++ b/classes/admin.php
@@ -745,7 +745,11 @@ class Admin
                 // Found the type and header from the session.
                 $data = $this->session->{$page->route()};
 
-                $header = ['title' => $data['title']];
+                // Initialise the page headers with the default page headers
+                $default_headers = $this->grav['config']->get('plugins.admin.editor.default_headers');
+                $header = Yaml::parse($default_headers);
+
+                $header['title'] = $data['title'];
 
                 if (isset($data['visible'])) {
                     if ($data['visible'] == '' || $data['visible']) {
--- a/classes/admin.php
+++ b/classes/admin.php
@@ -749,7 +749,7 @@ class Admin
 
                 // Merge the default page headers into $header
                 $default_headers = Yaml::parse(
-                  $this->grav['config']->get('plugins.admin.editor.default_headers'));
+                  $this->grav['config']->get('plugins.admin.frontmatter.default_headers'));
 
                 $header = (object) array_merge((array) $header, (array) $default_headers);
 
--- a/classes/admin.php
+++ b/classes/admin.php
@@ -745,11 +745,13 @@ class Admin
                 // Found the type and header from the session.
                 $data = $this->session->{$page->route()};
 
-                // Initialise the page headers with the default page headers
-                $default_headers = $this->grav['config']->get('plugins.admin.editor.default_headers');
-                $header = Yaml::parse($default_headers);
+                $header = ['title' => $data['title']];
 
-                $header['title'] = $data['title'];
+                // Merge the default page headers into $header
+                $default_headers = Yaml::parse(
+                  $this->grav['config']->get('plugins.admin.editor.default_headers'));
+
+                $header = (object) array_merge((array) $header, (array) $default_headers);
 
                 if (isset($data['visible'])) {
                     if ($data['visible'] == '' || $data['visible']) {
@@ -770,7 +772,12 @@ class Admin
                 }
 
                 if ($data['name'] == 'modular') {
-                    $header['body_classes'] = 'modular';
+                  // Make sure body_classes exists, so we can concat 'modular' to it
+                  if(!array_key_exists('body_classes', $header)) {
+                    $header['body_classes'] = '';
+                  }
+
+                  $header['body_classes'] .= ' modular';
                 }
 
                 $name = $page->modular() ? str_replace('modular/', '', $data['name']) : $data['name'];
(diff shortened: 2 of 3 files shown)


themes/grav/templates/forms/fields/editor/editor.html.twig
```diff
@@ -5,8 +5,10 @@
 
 {% block label %}
     {% if field.label %}
-        {% set hint = field.help ? 'data-hint="' ~ field.help|tu|raw ~ '"': '' %}
```
stephencookdev: This is crazy, and I have absolutely no idea what's causing it - if anyone more familiar with Twig knows, I'd be really interested in hearing!

I noticed the help text wasn't appearing correctly, and the HTML generated for the `data-hint` looked like `data-hint=""Default" headers to add to new pages&quot;`

It looks like Twig processes strings weirdly when surrounded in quotes.

For example, the code before this diff, and this following sample:

```twig
{{ "\"" ~ field.help ~ "\"" }}
```
if `field.help` is `"some test"`, will produce the output `""some" test&quot;`

Whereas

```twig
"{{ field.help }}"
```
correctly generates `"some test"`

No idea why... So this just side-steps the issue

w00fz: I think it should be `|e('html_attr')` instead of `|raw`

blueprints.yaml
```diff
@@ -321,4 +321,23 @@ form:
       label: Visitors history
       default: 20
 
+    Editor:
+      type: section
+      title: Editor
```
w00fz: `Frontmatter` would probably be more appropriate than `Editor`. Editor is very generic and doesn't really refer to the frontmatter only. The Content, for instance, is using the editor field, which would make this set of options a bit confusing.

blueprints.yaml
```diff
@@ -321,4 +321,23 @@ form:
...
+      underline: true
 
+    editor.default_headers:
+      classes: frontmatter
+      type: editor
+      label: Default Headers
```
w00fz: I would change this to `Default Page Headers`.

blueprints.yaml
```diff
@@ -321,4 +321,23 @@ form:
...
+      codemirror:
+        mode: 'yaml'
+        indentUnit: 4
+        autofocus: true
+        indentWithTabs: false
+        lineNumbers: true
+        styleActiveLine: true
+        gutters: ['CodeMirror-lint-markers']
+        lint: true
+      help: Default headers to add to new pages
```
w00fz: `Default page headers to automatically add to any new page created` ... Something along those line so it's more descriptive

classes/admin.php
```diff
@@ -745,7 +745,11 @@ public function getPage($path)
                 // Found the type and header from the session.
                 $data = $this->session->{$page->route()};
 
-                $header = ['title' => $data['title']];
+                // Initialise the page headers with the default page headers
+                $default_headers = $this->grav['config']->get('plugins.admin.editor.default_headers');
+                $header = Yaml::parse($default_headers);
+
+                $header['title'] = $data['title'];
```
w00fz: I'd prefer to have the title before the default page headers. It has become kind of standard to always have the `title` as the first entry of a frontmatter. It also feels more natural to have the user defined entries be after the core ones.

This could potentially also allow to override the title, even though it would be crazy to do so, there might be some scenarios out there, where this might be desired. 😄

themes/grav/templates/forms/fields/editor/editor.html.twig
```diff
@@ -5,8 +5,10 @@
...
-        <div class="form-label form-field hint--bottom" {{ hint }}>{{ field.label|tu|raw }}</div>
+        <div
+          class="form-label form-field hint--bottom"
+          {% if field.help %}data-hint="{{field.help|tu|raw}}" {% endif %}
+        >{{ field.label|tu|raw }}</div>
```
w00fz: You should try `|e('html_attr')` instead of `|raw` (but keep the `|tu` for translation.

```yaml
{% if field.help %}data-hint="{{field.help|tu|e('html_attr')}}" {% endif %}
```

blueprints.yaml
stephencookdev: Yeah, in my mind this "Editor" section could eventually have properties for all parts of the editor, including the Content part of it.

It seems a bit weird to me to have a config section specifically for a sub-part of a component, rather than for the whole component.

Happy to update if you're sure, though

themes/grav/templates/forms/fields/editor/editor.html.twig
stephencookdev: `|e('html_attr')` doesn't fix this particular Twig weirdness, I don't think - if that's what you meant

If you just meant my code, though, then yeah you're right - my bad over-eagerly copy-pasting

classes/admin.php
stephencookdev: Yeah good point. I've changed it so the default headers are merged in after the title is set

I've also made a slight change to how the `body_classes = modular` is set, so it's concatenated with whatever's set in `default_headers`

blueprints.yaml
w00fz: To me makes more sense being `Frontmatter` as this is what it refers to at the moment. If we expand in the future to include other settings then I agree we could update the terms but as it is now I find it less confusing if it was referencing to just the frontmatter.

Eventually we could give the section a name of "Pages" and something like "Default Frontmatter Page Headers" for the label.

stephencookdev: Fair enough :) - updated

w00fz:
Looks good to me. Cheers!

rhukster:
Will take a look today! promise!

classes/admin.php
```diff
@@ -747,6 +747,12 @@ public function getPage($path)
 
                 $header = ['title' => $data['title']];
 
+                // Merge the default page headers into $header
+                $default_headers = Yaml::parse(
+                  $this->grav['config']->get('plugins.admin.frontmatter.default_headers'));
+
+                $header = (object) array_merge((array) $header, (array) $default_headers);
+
```
rhukster: Ok, i'm going to propose something a bit radical.  I can see what you are doing and why it's in admin, but frankly I think we can improve this even more by breaking out this functionality out into a standalone **admin-default-frontmatter** plugin.  

What I mean by this is this code above would simple be an event along the lines of `onAdminCreatePageFrontmatter`, where we pass the default headers.  Then you would move your config and simple logic above to a simple plugin that listens to this one event. It would then return the merged headers.

I think this is a better approach because it allows more powerful functionality.  You could enhance your plugin to be a list of setups where you have wildcard routes with different page frontmatter.  So some defaults of any pages in `/blog/*` and another set of default frontmatter for `/news/*` for example.  Or Another plugin entirely that does some cleanup or formatting of titles. 

Anyway, it wouldn't be tied and limited to the admin, and would give you the freedom to develop your plugin as you wished without having to go through the admin plugin itself.

What this means is the following:

* I'll accept the fix below for the modular body classes.  That's a bug really and a good catch. 
* I'll add the event to make it possible to create a plugin doing exactly what you are already doing.
* You create a new plugin (we can help if you like) with the params and logic you already have in this PR.

What do you think?

stephencookdev: IMO it seems quite unnecessary to split this particular functionality into a separate plugin, as it seems appropriate as core admin functionality, to me.

The more powerful functionality you mention is a cool idea, but doesn't need this to be in a new plugin for it to exist.

But I don't know the codebase as well as you. In case you're adamant, I'll start drafting up a plugin to do this, and raise a separate PR do this repo to create the `onAdminCreatePageFrontmatter` event (unless you particularly wanted to), and we can compare the 2 solutions I guess?

rhukster: Funny, a request came up today on twitter for a plugin that is exactly the kind of thing that would benefit from this event, so I just added it and created the simple plugin.

Here's the commit that adds the event: https://github.com/getgrav/grav-plugin-admin/commit/1ea0a5490319e91609dc1b5e04243832595b3b2e

Auto Date Plugin: https://github.com/getgrav/grav-plugin-auto-date

You can use this plugin as an example to create your own.  Frankly it took me 10 minutes with the help of `devtools` plugin to create the basic plugin - https://github.com/getgrav/grav-plugin-devtools

rhukster:
So i'm closing this PR.  It was very useful however, as it served as the impetus to create a new event to allow plugins to work with the page header on creating a new page in the admin.

PR #812, opened 2016-10-06
fix problem when $parent is null

nogo:
When creating a page on root, there is no $parent object. The move function expect an non null object.

### DIFF
--- a/classes/controller.php
+++ b/classes/controller.php
@@ -1897,7 +1897,9 @@ class AdminController
             $original_order = intval(trim($obj->order(), '.'));
 
             // Change parent if needed and initialize move (might be needed also on ordering/folder change).
-            $obj = $obj->move($parent);
+            if ($parent != null) {
+                $obj = $obj->move($parent);
+            }
             $this->preparePage($obj, false, $obj->language());
 
             // Reset slug and route. For now we do not support slug twig variable on save.


flaviocopes:
`$parent` should never be null, as 

```php
$parent = $route && $route != '/' && $route != '.' ? $pages->dispatch($route, true) : $pages->root();
```

sets it to the parent page, or to the root page.

Do you have a specific problem I can replicate that led you to this PR?

nogo:
@flaviocopes  Okay you can reproduce with following steps

1. Login into Admin
1. Goto pages
1. Add a new page with following settings:
  - Page Title: Something
  - Page Folder: something
  - Parent Page: ROOT
  - Page Template: External
  - Visible: True
1. Now continue

flaviocopes:
I was able to reproduce it on the latest published release, but not on the `develop` branch, so it's already been fixed there. If you can test it (just download the latest package and put it under user/plugins/admin), please confirm you don't have the issue any more.

nogo:
Hi @flaviocopes I tried the develop branch and getting

Argument 1 passed to Grav\Common\Page\Page::move() must be an instance of Grav\Common\Page\Page, null given, called in /var/www/html/user/plugins/admin/classes/controller.php on line 1916

1. Login into Admin
1. Goto pages
1. Add a new page with following settings:
  - Page Title: Something
  - Page Folder: something
  - Parent Page: ROOT
  - Page Template: External
  - Visible: True
1. Now continue
1. Fill out the form with name and url
1. Save
-> Error happend

Did you habe another grav instance ? (My: Grav v1.1.5 - Admin v1.2.2 (git - develop))

flaviocopes:
You're using the develop branch of both Grav and Admin?

nogo:
only for Admin, I will check it with Grav in develop branch

flaviocopes:
Closing for inactivity

PR #814, opened 2016-10-09
Always delete image alternatives in AdminController#taskDelmedia

fredrikekelund:
Previously, image alternatives would only be deleted if the base image wasn't found, and only the first three alternatives would be removed. This commit changes that, so that all image alternatives are always deleted when the base image is.

### DIFF
--- a/classes/controller.php
+++ b/classes/controller.php
@@ -1288,11 +1288,9 @@ class AdminController
         }
 
         $fileParts = pathinfo($filename);
-        $deleteCount = 0;
+        $result = unlink($targetPath);
 
-        if (unlink($targetPath)) {
-            $deleteCount++;
-        } else {
+        if (!$result) {
             $this->admin->json_response = [
                 'status'  => 'error',
                 'message' => $this->admin->translate('PLUGIN_ADMIN.FILE_COULD_NOT_BE_DELETED') . ': ' . $filename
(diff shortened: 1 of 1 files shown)


classes/controller.php
```diff
@@ -1266,62 +1266,56 @@ protected function taskDelmedia()
...
+        if (unlink($targetPath)) {
+            $deleteCount++;
         } else {
             $this->admin->json_response = [
                 'status'  => 'error',
-                'message' => $this->admin->translate('PLUGIN_ADMIN.NO_FILE_FOUND')
+                'message' => $this->admin->translate('PLUGIN_ADMIN.FILE_COULD_NOT_BE_DELETED') . ': ' . $filename
             ];
+
+            return false;
```
fredrikekelund: Given the previous behavior, I'm not sure whether it's best to return here or to try and find and delete any image alternatives first. Some users may be stuck in a situation where they previously have tried to delete an image, and it wouldn't go away because it had an image alternative named `@4x` or higher. If we opt for this way of doing it, those files won't go away even if this patch is merged.

Probably an edge case, but maybe it wouldn't hurt to search for image alternatives in any case.

fredrikekelund:
Ping! Anyone had a chance to take a look yet?

flaviocopes:
Looks good to me

fredrikekelund:
Any news on this one? :)

fredrikekelund:
@flaviocopes done!

flaviocopes:
Thanks!

Issue #694, opened 2016-07-12
Invalid Security Token when attempting to save - Ms Edge


Potolozi:
When attempting to save page in Admin plugin I get an Invalid Security Token error message and page is not saved. I'm running the latest v1.1.0 rc3 version. My site lives on shared hosting.
I get this message when using Edge in Win10 on private network (not behind proxy).

1. I tried clearing browser cache – still the same error.
2. I changed to Chrome and performed the same action – no error.

It seems that it is Edge related.

w00fz:
Unfortunately this fix introduced more serious problems, forcing us to revert it.
I reopened and I will investigate further on how to address this better.

w00fz:
I think I nailed 🔨  it now!

maxfrigge:
This is still not working for me on a front end form. Is it possible that the fix only works for admin forms?

flaviocopes:
@maxfrigge on Edge? And just on that browser?

maxfrigge:
@flaviocopes Edge and Internet Explorer 11 on Windows10.

flaviocopes:
You also tried Chrome (for example) and it has no error?

w00fz:
I'm not sure how this is in any way related to frontend forms.
Frontend forms do not act like Admin, they are regular HTML5 forms and there is no JS magic behind the scenes, you submit regularly as you would do with any other HTML form.

In admin the issue was related to the fact that we have a complex form and we handle a lot via JS which caused issues with IE in some circumstances. Frontend forms and backend forms are completely unrelated though.

Please could you share screenshots/urls of what exactly is not working?

maxfrigge:
I will update the system to the latest Win10 and Edge Version, because they seem to be rather old. Once I am done with that I will test multiple browsers and report back. From what I have seen now the nonce is getting send to the server as expected. I am not saying that this is due to the "send twice issue", but I can say for sure that I was not able to submit either frontend nor admin forms due to nonce validation. 

Will get back with results asap.

maxfrigge:
I have upgraded the system and setup the web server with new website data. Dumped all caches etc.. here are the results.

In Chrome everything works just fine.

Frontend forms are working now (all browsers).

MS Edge keeps showing errors in Admin (Invalid security) as soon as I navigate to the Pages section. When I edit it a Page it does save the changes, but I also get the invalid security token error message.

maxfrigge:
[...]

PR #816, opened 2016-10-10
fix(ie): rename page move form to prevent blueprint submitting twice

maxfrigge:
Discussed in #694


flaviocopes:
Uhm.. Interesting find about this IE/Edge quirk. 

We also have the twigs under partials/blueprints-* that use the 'blueprints' form_id, but should not matter as they are not called in a modal beside another form.

flaviocopes:
Didn't test on Edge/IE yet, on Chrome all 🆗

Issue #848, opened 2016-11-03
Forgotten password page vulnerability


rsip22:
The forgotten password page also has a vulnerability. When you try and request a forgotten password link, the system confirms whether the e-mail is registered or not. I would suggest changing it to a more generic message to prevent this from being exploited.

flaviocopes:
PR merged

PR #849, opened 2016-11-03
A proposal to fix Issue #848

rsip22:
Removed the possibility to enumerate the users using the forgot password feature. Fixes issue #848


w00fz:
I agree this should be more generic and not show which username has been used.

PR #850, opened 2016-11-03
Add permissions field (don't merge yet)

flaviocopes:
screen shot 2016-11-03 at 18 59 16

Usage: add 

```
                    access:
                        type: permissions
                        label: Permissions
                        validate:
                            type: array
```

instead of

```
                    access.admin:
                        type: array
                        label: PLUGIN_ADMIN.ADMIN_ACCESS
                        multiple: false
                        validate:
                            type: array

                    access.site:
                        type: array
                        label: PLUGIN_ADMIN.SITE_ACCESS
                        multiple: false
                        validate:
                            type: array
```

in account.yaml or group.yaml

TODO:

- [x] Fix saving not set properties, currently they save as empty values instead of not saving at all. Update: DONE in https://github.com/getgrav/grav/commit/3838de1d97d8aee241caa9eddc17cb7e53d560c4


mahagr:
Is there a way to add my own permissions from a plugin?

flaviocopes:
Yes, through the already-existing `onAdminRegisterPermissions` event

Issue #864, opened 2016-11-16
add new `afterAdminSave` hook


amtur8:
I have first posted a question on forum: https://getgrav.org/forum#!/general:afteradminsave

I see this as a simple add in `classes/controller.php` (around line 2005) in _Admin_ plugin when the save is effective. I could create a Pull Request if it could help.

rhukster:
merged.. thanks.

PR #866, opened 2016-11-17
added 'afterAdminSave' hook

amtur8:
this event is fired just after the save was correctly completed from Admin plugin

references issue #864


amtur8:
I corrected the code to comply to your change request. 

Thanks a lot for your time!

PR #901, opened 2016-12-23
spaceless tag

Stepanov-Sergey:
Spaceless tag to remove whitespace between HTML tags. Slightly reduce the output code and faster loading


Stepanov-Sergey:
I agree, better to make the switch https://github.com/getgrav/grav/issues/1190

PR #916, opened 2017-01-08
Fix 'Disponibiles' => 'Disponibles'

filisko:


flaviocopes:
Translations for Admin are managed in https://crowdin.com/project/grav-admin, and just merged in here when we do a release - can you propose the edit there? Thanks!

filisko:
Ok, just suggested this change in crowding. :smile:

PR #922, opened 2017-01-12
Update section.html.twig

OleVik:
Close an open paragraph-tag, thus avoiding incorrect HTML and an empty extra paragraph.


flaviocopes:
👍

PR #924, opened 2017-01-12
Conditional check for label in field.html.twig

OleVik:
If a blueprint-defined field does not have a label set, do not render the `.form-label`-structure (`<div class="form-label"><label></label></div>`) as the `label`-element's padding represent a small but unnecessary piece of empty space.

### DIFF
--- a/themes/grav/templates/forms/field.html.twig
+++ b/themes/grav/templates/forms/field.html.twig
@@ -11,29 +11,31 @@
 {% block field %}
     <div class="form-field grid{% if vertical %} vertical{% endif %}{% if field.toggleable %} form-field-toggleable{% endif %}">
         {% block contents %}
-            <div class="form-label{% if not vertical %} block size-1-3{% endif %}">
-                {% if field.toggleable %}
-                    <span class="checkboxes toggleable" data-grav-field="toggleable" data-grav-field-name="{{ (scope ~ field.name)|fieldName }}">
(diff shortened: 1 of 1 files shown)


OleVik:
Though an argument can be made that the label should be required, this `label` does not necessarily relate to an `input`, therefore it can be extraneous.

filisko:
Not a bad idea ...

flaviocopes:
The same UI can be achieved using `style: vertical` in the field, which is used to put the label on top instead of left-side. Without label, the field simply takes 100% width.

Keeping things as-is allows to have fields with no label stay in the right column. I would not change it.

OleVik:
Using the following, with `style: vertical` active:

```
    header:
      type: section
      title: Theme
      underline: false
      text: Select a theme for the admin interface by clicking your preferred appearence, and then click the "Save"-button.
      fields:
        current:
          type: radioimage
          classes: fancy
          name: current
          style: vertical
          default: grav_admin
          options:
            grav_admin: Default
            metal: Metal
            light: Light
            starkcontrast: Stark Contrast
            tan: Tan
            grey: Grey
```

I get the aforementioned markup where the label has a 5px padding on the bottom. Hardly massive or perhaps even significant, but superfluous as an empty element. As said though, apart from the small, empty element there is no particular reason for removing it.

Closing as an unnecessary fix.

PR #973, opened 2017-02-13
Admin panel cleaner look for tabs, theme & plugin

coder4life:
Improvement to tab parts of admin interface.

## Ideas:
* One idea was to inject extra HTML code higher up in the twig structure into the field tab interface. However this would require the storage of context to be passed to each twig instance, complicating the structure
* Second idea was to inject this on JavaScript side. However that attributes to more DOM traversal and manipulation, and performance is better to render this server-side.
* Third idea was to have special config parameters added to the actual tab interface. However this would not make the tab instance generalized

## Proposed:
Custom field type for themes and plugins. Conditional context awareness if this is a plugin or a theme. The theme or plugin twig variables only exist in the blueprint context is appropriate. If in theme, the plugin twig variable is null. If in plugin, the theme twig variable is null. We can use this to check the field appropriately and output a custom message so developers can know what field type to use when.

## Backward compatibility:
Should be no issue. Code will default behavior unless specified in blueprint code.

### DIFF
--- a/languages/en.yaml
+++ b/languages/en.yaml
@@ -636,3 +636,4 @@ PLUGIN_ADMIN:
   OFFLINE_WARNING: "The connection to the GPM cannot be established"
   CLI_COMPATIBILITY: "CLI Compatibility"
   CLI_COMPATIBILITY_HELP: "Ensures that only non-volatile Cache drivers are used (file, redis, memcache, etc.)"
+  FIELD_CONTEXT: "Wrong field type, only shown when the blueprint context is part of the"
--- a/themes/grav/templates/forms/fields/plugininfo/plugin.html.twig
+++ b/themes/grav/templates/forms/fields/plugininfo/plugin.html.twig
@@ -1,0 +1,13 @@
+{% extends "forms/field.html.twig" %}
+
+{% block input %}
+
+    {% if plugin is not empty %}
+        {% include 'partials/plugins-info.html.twig' %}
+    {% else %}
+        <div>
+            <p>{{ "PLUGIN_ADMIN.FIELD_CONTEXT"|tu }} {{ "PLUGIN_ADMIN.PLUGIN"|tu }}</p>
+        </div>
+    {% endif %}
+
+{% endblock %}
--- a/themes/grav/templates/forms/fields/themeinfo/themeinfo.html.twig
+++ b/themes/grav/templates/forms/fields/themeinfo/themeinfo.html.twig
@@ -1,0 +1,13 @@
+{% extends "forms/field.html.twig" %}
+
+{% block input %}
+
+    {% if theme is not empty %}
+        {% include 'partials/themes-info.html.twig' %}
+    {% else %}
+        <div>
+            <p>{{ "PLUGIN_ADMIN.FIELD_CONTEXT"|tu }} {{  "PLUGIN_ADMIN.THEME"|tu }}</p>
+        </div>
+    {% endif %}
+
+{% endblock %}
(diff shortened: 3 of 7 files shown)


rhukster:
Is there a screenshot of what this changes?  if it's purely just a code change, can you provide a use case or test plugin that uses it.  Cheers!

coder4life:
Normally there is no way to separate out the information as part of the tab interface. Although not a huge problem, if you do have a lot of information in the default theme it pushes many of the tab and config operations lower. This can introduce a scroll and click operation. Although minor visually, this was an attempt to get the interface to look consistent to standard blueprints interface for page types. The way it was implemented was to prevent changing the default behavior, but allow tabs as an option for displaying the information. Upon first look it should always start with info as long as it is the first tab. The goal was to allow this for the plugin and theme interface.

## Before change:
screen shot 2017-02-12 at 6 34 18 pm

## After change:
screen shot 2017-02-12 at 6 35 16 pm

coder4life:
There is a minor display bug here as well. If you notice the header is part of the field not a direct child of the necessary CSS class name. As such the section header underline does not extend full width. I am not sure the best way to fix this, other than having a special embed condition.

This might be able to be fixed in a few ways, but maybe someone has a better idea.

w00fz:
I'm not quite sure how to enable this, I have a plugin with tabs and your changes don't switch to top tabs like in your screenshot.
It looks like in order to get this behavior you have to make some definitions in the blueprint that appear a bit obscure. I'm also finding this check kind of flaky `{% if(plugin.form.fields.tabs.fields.info.fields.id.type != 'plugininfo') %}`.

I can see having the info omni-present be annoying and taking up too much room but I don't like the idea of turning that into a field. As a field it can too easily be overridden and changed. I think we should keep that portion of plugins and themes consistent throughout the admin even if it means being opinionated about it.

I would say a better approach for this could be either:
1. Make all plugins and themes info show as an Info tab like you are wanting it to be. This would maintain consistency and I can see it being an enhance. 
2. Keep the current behavior but make the whole info data collapsible like it is in the list of plugins. For those plugins where there's a lot of fields, the user can collapse the whole info details and concentrate on the fields.

coder4life:
`if(plugin.form.fields.tabs.fields.info.fields.id.type != 'plugininfo') %}`

I agree this is less than ideal. However there is something very similar, not my code, doing something very similar a few lines down in the file. I can change that to a filter function instead to look for at least first occurrence of the key.

> 1. Make all plugins and themes info show as an Info tab like you are wanting it to be. This would maintain consistency and I can see it being an enhance.
Please look at my Ideas section above to list the different approaches this could be achieve and associated caveats. It is possible to make it show as a tab, but how the twig and extension is currently structured it is a bit difficult to pull off without changing a lot of code. That assumption being that tabs have to be adjacent siblings.

Goal was this to be not intrusive to current setups. Also `form/fields/userinfo.hmtl.twig` field does something very similar as the added files.

> 2. Keep the current behavior but make the whole info data collapsible like it is in the list of plugins. For those plugins where there's a lot of fields, the user can collapse the whole info details and concentrate on the fields.
Another idea I considered too, however has its own issues.  What if someone returns, do we store something extra too keep it closed. Do we require someone to close it everytime.

A final option is to remove fields completely in that section and have a specialize page just for configurations. Maybe a gear icon on the theme selector page.

w00fz:
We already keep a state for tabs, it gets stored when you change tab and restored on load. The same would apply for both of my propositions above, whether it was a tab or a collapsible panel.

Removing the info completely I don't think is an option.

coder4life:
I meant the tabs would not be adjacent how they are set up. Kind of breaks semantics a bit.

- Tab 0
- Tabs
- - - Tab 1
- - - Tab 2
- - - Tab 3
- - - Tab 4

I meant moving the forms to another page/panel. Not removing the info at all. The info page would be the same

rhukster:
I have a better solution.. new Tabs with `subtle` class on the tabs field:

2017-03-10 at 4 44 pm

Issue #965, opened 2017-02-11
Cannot scroll after update


ivapelc:
I have fresh install of Grav + Admin. After todays upgrade of both Grav and Admin Plugin, I cannot scroll in the backend (no scrollbars)...

rhukster:
We bet you are using cloudflare? https://github.com/getgrav/grav-plugin-admin/issues/808#issuecomment-276824595

You should be able to disable an option as outlined by this other thread.

ivapelc:
I'm not - at least don't know about it. It worked just fine before the update.

rhukster:
Try clearing your browser cache or try another browser. Actually nothing changed in this release that would change that behavior.  

--  
Andy Miller

rhukster:
What's the URL of your site?

rhukster:
Also, try the latest version that was just released with a JS bundle rebuild.  Should sort it.

ivapelc:
Reinstalled whole Grav + Admin and everything worked fined until I switched language to "Czech". Then the problem occured again (no scroll; no Maintenance status on Dashboard; error after performing Check for updates). After switching back to "English" and clearing cache, it seems to work fine again, so the problem seems to be related to Czech translation...

flaviocopes:
@ivapelc quick fix - edit user/plugins/admin/languages/cs.yaml like in here https://github.com/getgrav/grav-plugin-admin/pull/974/files

I fixed this in the Admin plugin for the next release.

PR #974, opened 2017-02-13
fixed issue with non-escaped quotation marks in CS language [issue #965]

Miclik:
reported in #965

it caused error in JS evaluation, so page scrolling didn't work


flaviocopes:
Language files must be changed in Crowdin: https://crowdin.com/project/grav-admin 

It will be merged in the next release.

Can you apply the change there? Thanks!

flaviocopes:
(I replicated the problem this issue solved. Fixed in Admin instead than in the language file, which is correctly formatted)

flaviocopes:
https://github.com/getgrav/grav-plugin-admin/commit/799f13bdfb854e818ebcfbc22dc1ba4f5af9d16e

PR #983, opened 2017-02-18
Fixed path separator for PHP on Windows

zzzzBov:


rhukster:
Good catch!

w00fz:
Oops, I missed that.
Before re re-release this needs js recompilation.

PR #991, opened 2017-02-21
Separate userinfo avatar and avatar credit in separate twig

OleVik:
Follow-up on #989, modularizing avatar (and its credit-line) for the userinfo-field in admin/user/ as well.


OleVik:
Reverted the range-field commit, that's supposed to be a different PR.

PR #1015, opened 2017-03-12
Allow full path render of pages in selector.

coder4life:
Full path is off by default.


coder4life:
To give some context. With small sites this is not much of a problem, but once you have up to 5 levels of nested structure that could have similar names at various different child depths. The current indent method becomes hard to digest.

Issue #786, opened 2016-09-19
Page Ordering Disabled


metadiv:
I cant seem to get the admin ordering fields to work for pages in a collection of prefixed folders.

**Page Order**
Parent setting order, ordering disabled

appears above the stack of page name 'advanced' tab.

![image](https://cloud.githubusercontent.com/assets/1914218/18640708/3897f7fe-7e91-11e6-9871-020e39e81f16.png)

In the parent page's markdown file I am using :

``` markdown
---
title: 'page title'
content:
    items: '@self.children'
    order:
        by: default
        dir: asc
---
```

Previously this worked but I have recently upgraded my site to the latest version of Grav and Admin.

I can confirm that manually modifying the folders prefixes does re-orders them correctly in the front end, but as I need to hand this site off to less techy colleagues it would be great to give them control from the admin ui.

Any suggestions please?

rhukster:
This looks like a bug.  Will get it looked at.  Thanks.!

metadiv:
@rhukster Yes it worked on the previous build I had before I upgraded.

One other question. Do you know if its possible to setup either the entire system or via a page type blueprint to automatically enable folder prefixing even if the page created is the first sibling in its parent?

rhukster:
Ok, there is a bug here, but it's still not going to work for you until you remove the `items:` section of the frontmatter.  This is because you cannot manually reorder if you are setting the `content.items` manually. Manual reorder is just for an arbitrary collection of page folders in a particular directory: https://github.com/getgrav/grav-plugin-admin/blob/develop/themes/grav/templates/forms/fields/order/order.html.twig#L17

Imagine if you had a ` content: items: '@taxonomy....` collection.  This cannot be reordered of course.

This has not changed since Feb 1st.  There was a bug where this logic was not being tested properly.  Perhaps that is the version you updated from? If so it was a buggy version, that is now fixed.

metadiv:
@rhukster Thanks for the explanation. I can confirm removing the item: self.children from the parent works. 

Can I ask though my 2nd question, is there a way to enable folder prefixing as a default, either int he site settings or as a blueprint function?

rhukster:
Please open a new issue for this question. Thanks!

PR #1026, opened 2017-03-17
don't disable page ordering when the collection is @self.children and the order is default

benface:
This fixes the undesirable behaviour of disabling the ordering of a page's children when it defines a collection of items with `header.content.items: @self.children` and `header.content.order.by: default`. In that case, the collection on the front-end is effectively in the order set by the user in the admin panel (which is also the folders prefix order), so it makes sense to enable it and there shouldn't be any confusion like there is when the collection is not the page's children, as explained to me by @rhukster on Slack.

Issue #786 was about this. It was closed because a workaround was suggested, but I think this fix is better as it allows you to use some of the great features of collections such as `url_taxonomy_filters` to filter through the page's children to display, in addition to letting the admin set their order.

### DIFF
--- a/themes/grav/templates/forms/fields/order/order.html.twig
+++ b/themes/grav/templates/forms/fields/order/order.html.twig
@@ -14,23 +14,27 @@
     </div>
     <div class="form-data block size-2-3 pure-u-2-3">
         <div class="form-order-wrapper {{ field.size }}">
-            {% set canReorder = not data.parent.header.content.items and data.visible %}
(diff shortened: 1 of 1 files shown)


rhukster:
Actually we have a refactoring effort underway to radically change how reordering works for pages.  It will focus on numeric prefixes and leave collections completely out of the equation.  We also plan to allow you to reorder all siblings in one shot rather than one at a time as you have to now.

PR #1054, opened 2017-04-08
RTL Support for admin plugin

qasedak:
I've added some RTL styles and Fonts for Arabic, Persian and Hebrew languages. I'll check RTL styles and improve them by time.
Grav admin RTL
TnX.


rhukster:
@qasedak do you mind updating your PR with latest changes in admin plugin so i can test with latest?

mahagr:
If someone can update this one to be par with Grav 1.7, we could consider this addition. But as it is, we cannot merge this.

qasedak:
I'll give it a try 😉

mahagr:
@qasedak Thank you!

qasedak:
I've decided to make a fresh fork and also a new Pull request from it as this one is too old...
you find it here:
https://github.com/qasedak/grav-plugin-admin/tree/rtl-support

mahagr:
Btw, you can use `language_codes.rtl(code)`...

mahagr:
I'm also wondering, should RTL be activated when the current language in the page is set to RTL language. You can get the current language by `grav.language.language ?? 'en'`, so you could use `language_codes.rtl(grav.language.language ?? 'en')`.

qasedak:
yes sure! thank you. I thought it's the user's language that will be displayed so if user switches to a RTL language  we should add the RTL.css style...
but why `grav.language.getActive`  doesn't work? as [documented here](https://learn.getgrav.org/17/content/multi-language#language-logic-in-twig-templates)

qasedak:
> 
> 
> I'm also wondering, should RTL be activated when the current language in the page is set to RTL language. You can get the current language by `grav.language.language ?? 'en'`, so you could use `language_codes.rtl(grav.language.language ?? 'en')`.

@mahagr 
as I've checked the user's preferred language overrides the first supported language (admin config section) in Admin panel i used the code below: 
`{% if language_codes.rtl(grav.user.language) %}
    {% do assets.addCss(theme_url~'/css/rtl.css') %}
{% endif %}`

mahagr:
Ok, that will work. I guess it works fine in admin.

Issue #722, opened 2016-07-31
Page Editing in Fullscreen


jacobm001:
When using the page editor in the admin console, I've run into an issue trying to maximize the editor. If the browser tab isn't full screened, clicking the fullscreen option pushes the bottom content off screen. If the browser is maximized, the entire display is visible, but it still doesn't look quite correct.

When the browser tab is not full screened: 
![fullscreen](https://cloud.githubusercontent.com/assets/2245023/17280230/c2b1cbaa-573f-11e6-8aa9-8b4f4a3491c2.PNG)

When the browser tab is full screened:
![chrome_fullscreen](https://cloud.githubusercontent.com/assets/2245023/17280232/d242e1e4-573f-11e6-9439-d4f8cd9a6972.PNG)

I've duplicated the issue on two different Windows machines (one on 7, and one on 10) using the latest versions of Firefox and Chrome. I have not checked IE or Edge.

flaviocopes:
Looks like a Windows-only issue, cannot replicate on OSX.

jacobm001:
I've also noticed the issue on ChromeOS, so it doesn't appear to be Windows exclusive.

benyaminl:
Also on Linux with Chrome Canary 54.0

asirco:
I've noticed that when editing a page’s content and you put it in Full Screen mode and paste in some text, if the text goes past a certain physical height, the text area input stops. Truncating what’s at the bottom of the article so you can’t get to it and edit. It works perfectly fine when not in full screen mode - which kind of defeats the point of having a full screen mode.

flaviocopes:
Replicated in Chrome on OSX. The last line is not visible.

asirco:
For me, it happens in Chrome and Safari on OSX and Chrome and Firefox on Windows 7 Ultimate, and it's often the last few paragraphs, not just the last line. This is especially so on pages with 800+ words.

flaviocopes:
Fixed by https://github.com/getgrav/grav-plugin-admin/pull/1065

PR #1065, opened 2017-04-13
Fix fullscreen editor height #722

david-szabo97:
Hi there!

This should fix the issue with the fullscreen editor.

https://github.com/getgrav/grav-plugin-admin/issues/722

P.S. Whenever I tried to compile SCSS it generated to totally different source map files. I don't want to clutter the pull request with those.


flaviocopes:
Thanks! Tested 👍

PR #1067, opened 2017-04-14
More robust themes style

Sommerregen:
This PR address a CSS styling issue, see issue before my changes:

before

And after my changes:

after

I made changes to the SCSS files. Thus, a recompilation is necessary.


flaviocopes:
Tested and works 👍

PR #1096, opened 2017-04-26
Added an option for opening frontend pages in a new tab, in a separate tab (always the same) or in the current tab.

parcand:
This new option gives you more control over the "eye" button behaviour. This is particularly useful to users complaining about the fact that a new tab is created every time you click on the button.

capture d ecran 2017-04-26 a 16 57 36
capture d ecran 2017-04-26 a 16 56 29

### DIFF
--- a/admin.yaml
+++ b/admin.yaml
@@ -22,6 +22,7 @@ session:
 warnings:
   delete_page: true
 edit_mode: normal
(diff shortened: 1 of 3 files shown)


rhukster:
Like it! Thanks.

PR #1101, opened 2017-04-28
Bugfix: Save in wrong lang when multiple tabs open

dctremblay:
English and french in this PR are used only for example purposes.

When you have multiple tabs opened in different languages, when you save, it often cause to overwrite the french contents with the english contents (or vice-versa, or according to your specific language configuration).

How to reproduce the bug :

1. Open a tab and edit a page in french
2. In another tab, open the same page in english then save
3. Return to the french tab and save
4. **The english contents got overwrited by the french contents**

This patch makes the use of the language code in URL instead of the session when editing a page, preventing to overwrite the contents of another language while power-editing with many tabs in different languages.

### DIFF
--- a/admin.php
+++ b/admin.php
@@ -585,9 +585,8 @@ class AdminPlugin extends Plugin
         // Initialize Admin Language if needed
         /** @var Language $language */
         $language = $this->grav['language'];
-        $this->grav['session']->admin_lang = $language->getActive();
-        if ($language->enabled() && empty($this->grav['session']->admin_lang)) {
-            $this->grav['session']->admin_lang = $language->getLanguage();
+        if ($language->enabled()) {
+            $this->grav['session']->admin_lang = $language->getActive();
         }
 
         // Decide admin template and route.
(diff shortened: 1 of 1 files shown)


dctremblay:
@flaviocopes Ready to re-review

rhukster:
I had to revert this because it was losing track of the current language. If you switched to italian in pages list, then click on the italian page, it loaded in default lang (english).

dctremblay:
@rhukster What do you think is the more depressing between :

* Save a page and get switched to the language of this page
* Save a page and overwrite the other language of this page, **losing data**

Thanks !

rhukster:
Well I agree this needs to be fixed, but the fix actually caused more problems because it broke behavior.  When you were in a language and clicked a page, it loaded up in the default language! This is not an acceptable solution for an issue that occurs only when editing the site in multiple tabs (which is not common practice)

I'm definitely open to a fix that solves the issue and doesn't break things though.

dctremblay:
@rhukster Thanks for your answer, I agree with you. I will make further researches during this week in order to find a solution that fully resolve all of those issues.

Thank you very much !

Issue #1066, opened 2017-04-13
Custom add page options


jwillp:
As I explained in getgrav/grav#421, it would be great to have the possibility to add custom page types based on blueprints in the Add Page Menu.

For example the menu by default has these choices:
- Add Page
- Add Folder
- Add Modular
I'd like to be able to add thins like:
Add Blog Post
Add Gallery
Add Other special page type.

These options would automatically set things up so that a page for example would be created inside the blog/ directory with an "Item" page template in order to create a blog post etc..

This would be a very useful feature, especially when handing the site to clients who would not have to worry about page templates and folder structure.

fireraccoon:
I would need this soon for a client
do you have any indications on how I could implement this myself?

ricardo118:
Well when you tell a client to add a page, all they gotta do is select a Template for the right thing. If you set up the blog, the gallery, etc it will be just as easy. However, if you really wanted to add buttons, just duplicate the add page button but put the field template preselected and hidden (if you so wish)

fireraccoon:
@ricardo118 I wish i could simply tell them to select the right template and the right folder, unfortunately because of the way the modal is currently designed, the selection box of the folder lists all the pages of the site, which is problematic because in order to find the right folder they need to scroll through potentially hundreds of pages. Hence the predefined modals.

I was hoping for a more elegant solution to implement this than simply copy & pasting the modals, and directly editing the plugin's files. (Because of updates)
I did try to do it though, but I run into problems when trying to specify the default value of select boxes: Strangely enough I never get the right option by default.

Is there any way to extend the admin plugin without changing its code to do what I want ?

david-szabo97:
I think we could make a configuration for this.
I have this in my local version if we can agree on the configuration structure I can refactor it into a PR.

By default, it uses the template used by Add Page.

admin.yaml
```yaml
add_modals:
  post:
    label: Add Post
    blueprint: admin/pages/new_post
    template: custom_template
    link_classes: some_class
    modal_classes: some_modal_class
    with:
      some_data: for the template
    show_in: dropdown
  image:
    label: Add Image
    blueprint: admin/pages/new_image
    show_in: bar
```

An example blueprint. On my blog I have more than 1 authors so with the getNewPostRoute I can generate a route for the user, for example for my account it generates: "/dave/posts"

new_post.yaml
```yaml
form:
  validation: loose
  fields:
    section:
        type: section
        title: Add Post

    title:
      type: text
      label: Post Title
[...]

PR #1104, opened 2017-04-30
Feature custom page options

david-szabo97:
Related issue: https://github.com/getgrav/grav-plugin-admin/issues/1066

To avoid using JS, I added a custom value handling to 'folder' field: `@slugify-field`
If you want to automatically slugify the title and use it as your folder, just put `default: @slugify-title` to the folder field.

Possible values for `add_modals`:

- **label** - Text to be shown in the button
- **show_in** (default: bar) (values: bar|dropdown) - Determines whether to show the button in the **bar** or **dropdown**
- **blueprint** - Form blueprint to pass to modal template
- **template** - Template to use for modal (default: partials/blueprints-new.html.twig)
- **with** - Extra data for template
- **link_classes** - This will be added to the link's (button in bar) class attribute
- **modal_classes** - This will be added to the modal's class attribute

## Example
This example adds an "Add Post" button which only accepts one value, the title. Automatically puts it into the `/posts`. (You can find more examples and explanation in the https://github.com/getgrav/grav-plugin-admin/issues/1066)

**user/blueprints/admin/pages/new_post.yaml**
```yaml
form:
  validation: loose
  fields:
    section:
        type: section
        title: Add Post

    title:
      type: text
      label: Post Title
      validate:
        required: true

    folder:
      type: hidden
      default: '@slugify-title'

    route:
      type: hidden
      default: /posts

    name:
      type: hidden
      default: 'post'

    visible:
      type: hidden
      default: ''

    blueprint:
      type: blueprint
```

**user/config/plugins/admin.yaml**
```yaml
add_modals:
  post:
    label: Add Post
    blueprint: admin/pages/new_post
    show_in: dropdown
```

### DIFF
--- a/classes/admincontroller.php
+++ b/classes/admincontroller.php
@@ -655,7 +655,12 @@ class AdminController extends AdminBaseController
         }
 
         $route  = $data['route'] != '/' ? $data['route'] : '';
-        $folder = ltrim($data['folder'], '_');
+        $folder = $data['folder'];
+        // Handle @slugify-{field} value, automatically slugifies the specified field
+        if (substr($folder, 0, 9) == '@slugify-') {
+            $folder = \Grav\Plugin\Admin\Utils::slug($data[substr($folder, 9)]);
+        }
+        $folder = ltrim($folder, '_');
         if (!empty($data['modular'])) {
             $folder = '_' . $folder;
         }
--- a/classes/utils.php
+++ b/classes/utils.php
@@ -35,4 +35,26 @@ class Utils
         // If a User with the provided email cannot be found, then load user with that email as the username
         return User::load($email);
     }
+
+    /**
+     * Generates a slug of the given string
+     *
+     * @param string $str
+     * @return string
+     */
+    public static function slug($str)
+    {
+        if (function_exists('transliterator_transliterate')) {
+            $str = transliterator_transliterate('Any-Latin; NFD; [:Nonspacing Mark:] Remove; NFC; [:Punctuation:] Remove;', $str);
+        } else {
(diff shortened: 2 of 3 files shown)


flaviocopes:
Cool! 

Two considerations: 

1. can we use something else instead of cocur/slugify? Seems to add 3800+ files, which is a bit too many for such a simple thing as converting to a slug? Maybe add our own slugify functionality in https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Utils.php 

2. can you move the modal key in the modal itself, so it can be defined in a list in the Admin plugin blueprint? This way modals can be configured via the Admin panel.

We can then also make it possible for plugins to add new ones dynamically, with an event.

david-szabo97:
@flaviocopes 
To be honest I totally removed the key stuff. I think we can freely rely on the index. What do you think?

david-szabo97:
@flaviocopes Is there anything else I should do before this gets merged? Thank you.

flaviocopes:
I tested and it looks good! Can you move the PR desc to a Cookbook entry? So people know how to use this in their sites or also plugins https://learn.getgrav.org/cookbook

mahagr:
Umm, honestly I'm not really liking the change as it only applies to admin plugin and is hardcoded to the controller like that. I guess it's only a save thing, but still...

flaviocopes:
I like offering plugins, or also themes, the option to add custom page creation modals, as it's easier for clients.

What in particular you don't like @mahagr, the concept of some part of the code that's implementing this?

mahagr:
As I said, the `@slugify-title` only works from the admin when saving the data. Generally these processors are used inside of the `BlueprintForm` class to run a function call on `BlueprintForm::init()` call.

Issue #1022, opened 2017-03-16
Feature Request: Edit Next


frumbert:
In the admin toolbar there's a Back link. I'd like a "Next" link that leaves me at the edit screen but takes me to the next page in the current sequence, so that I can easily edit a bunch of pages in sequence without having to do the back-and-forth link. Handy on documentation-type templates with hundreds of pages I need to edit.

PR #1112, opened 2017-05-09
Add previous and next button when editing page

david-szabo97:
https://github.com/getgrav/grav-plugin-admin/issues/1022


rhukster:
My only concern with this is that the toolbar is already quite full, and this is going to make it even busier.  How does this degrade to mobile?

david-szabo97:
Yeah, unfortunately, we don't have too much space there.

nbcvbnc

rhukster:
I guess we could hide prev/next in mobile views.  They are not critical functionality.

Issue #956, opened 2017-02-03
Add a 'Limit' option for List Fields


ItsMeAra:
When using [The List Field](https://learn.getgrav.org/forms/blueprints/fields-available#the-list-field) it would be nice to be able to set a limit on the number of Items that can be added through the admin. For example, I tried `limit: 1` but didn't do anything, still allows unlimited items. 

```yaml
header.hero_button:
  name: button
  type: list
  label: Button
  limit: 1  # <~ this would be great!
  fields:
    .text:
      type: text
      label: Button Text
      validate:
        required: true
    .url:
      type: text
      label: Button URL
      validate:
        required: true
    .theme:
      type: select
      label: Button Color
      options:
          1: 'Blue'
          0: 'Green'
```

Discovered in Slack chat that this is not a current feature. Would be nice to add. 😄

dimitrilongo:
Same here https://github.com/getgrav/grav-plugin-admin/issues/720

w00fz:
Fixed in #1113, thanks @david-szabo97

PR #1113, opened 2017-05-09
Add min and max to list field

david-szabo97:
https://github.com/getgrav/grav-plugin-admin/issues/956


w00fz:
Nice addition! Thank you!

PR #1116, opened 2017-05-17
Remove save button and save location notification when filename isn't…

AvantaR:
Save button and "save location" notification bar are unnecessary on info tab in configuration.

Best regards


flaviocopes:
Definitely!

PR #1123, opened 2017-06-01
Setup for continuous integration using CircleCi

karfau:
I added `circle.yml` (and an additional `test-ci` script in `composer.json` that creates test reports) to be able to integrate with [CircleCI](https://circleci.com/). (The only reason for choosing this over others is that I have some experience with it.)

Activating it is as easy as signing up using github, and selecting which projects to build: 
https://circleci.com/docs/1.0/getting-started/

The result is that each push to github triggers a build over there and pings back the result to github, so that little icons appear beside the commits: 
https://github.com/karfau/grav-plugin-admin/commits/feature/circle-ci

An additional possible step is to only allow merges to a branch when all tests pass, its called protected branches and can be configured in the settings section of a repository.

It would also be possible to add front end tests for the grav theme, but in the `package.json` there are no tests configured, so I left it out.

I can also do the same thing for Grav itself, if wanted.

CircleCI is currently in beta phase for their version 2, but I don't have any experience with it and signup is currently more complicated. I'm happy to try it if wanted.

Best Christian

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ themes/grav/js/admin.js
 themes/grav/js/vendor.js
 themes/grav/js/*.map
 /.idea
+*.iml
 
 tests/_output/*
 tests/_support/_generated/*
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,9 @@
+machine:
+  php:
+    version: 5.5.31
+dependencies:
+  override:
+    - composer update
+test:
+  override:
+    - composer test-ci && mv tests/_output $CIRCLE_TEST_REPORTS/unit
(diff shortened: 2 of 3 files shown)


karfau:
Not sure if this URl is public, but this would be the build result:
https://circleci.com/gh/karfau/grav-plugin-admin/4

karfau:
Sure enough, didn't look into the core project yet. 
I think enabling such an integration can also encourage people to add tests, because they matter.

As I have no experience with travis yet, I'm not sure I can contribute as much here.
You could of course support both travis and CircleCI config, and each fork could then decide how to do the CI, but this has also some overhead of maintaining both solutions.

I assume you have somebody on the team who can do the Travis integration as easily as I did the CircleCI one, So I won't invest additional time into this one, but if you want me to look into it,
ping me.

Otherwise feel free to pclose the PR.

Issue #1129, opened 2017-06-07
[First time launching Gravy] Register admin form does not give clues about what's wrong with wrong email


BoQsc:
I did type "admin@localhost" and registration form didn't give any clues about what's wrong with it.
A thing I thought about: maybe the password isn't matching before I arrive at a conclusion that email was incorrectly typed.

flaviocopes:
Thanks for reporting, there is HTML frontend validation in the form, but `admin@localhost` passes that. It does not pass the server-side validation, but the error message does not appear. Looking to improve that.

PR #1132, opened 2017-06-08
If a form is already defined, do not override it

flaviocopes:
Preserves form validation re
https://github.com/getgrav/grav-plugin-admin/issues/1129


rhukster:
Actually after testing, this fine as-is.  cheers @flaviocopes !

Issue #1172, opened 2017-07-21
'onAdminCreatePageFrontmatter' emits too few header information


THoenig:
Hi there.

I'd like to modify/create several fields on page creation with help of a custom page creation modal. As far as I can see, only 'title' and (if exists) 'visible' are getting emitted with the event in [Admin::getPage](https://github.com/getgrav/grav-plugin-admin/blob/develop/classes/admin.php#L1396)

Custom fields are simply ignored. Is this by intention, and if yes, why? From my point of view it makes very little sense as it is now.

rhukster:
Please check out my comments.

PR #1175, opened 2017-07-24
Make all existing information available on page creation

THoenig:
This is a possible solution for  #1172.

### DIFF
--- a/classes/admin.php
+++ b/classes/admin.php
@@ -1393,7 +1393,8 @@ class Admin
                 $page->name($name . '.md');
 
                 // Fire new event to allow plugins to manipulate page frontmatter
-                $this->grav->fireEvent('onAdminCreatePageFrontmatter', new Event(['header' => &$header]));
+                $this->grav->fireEvent('onAdminCreatePageFrontmatter', new Event(['header' => &$header,
+                        'data' => $data]));
 
                 $page->header($header);
                 $page->frontmatter(Yaml::dump((array)$page->header(), 10, 2, false));


rhukster:
This looks pretty good to me at first glance. I need to test it a bit though before merging.

rhukster:
Ok i'm seeing an issue with this PR when creating a new page using the + button:

2017-07-25 at 7 09 pm

These values are being set by the page create modal, but really they should not be set in the frontmatter at all, they are set as properties on the page, which in turn are used to set the values, but only title should be actually stored in the frontmatter.  This is why we had that mechanism to create a simple header with only title (and possibly 'visible')

THoenig:
I think i found a simple solution that keeps the default process untouched but adds the ability to handle additional fields.

@rhukster Could you please take a look?

rhukster:
Yup, that works for me :)  👍

justinhartman:
@THoenig or @rhukster can one of you tell me how I pass frontmatter options to the `'data' => $data` object so that it is saved in the frontmatter on page creation as per [this screenshot](https://github.com/getgrav/grav-plugin-admin/pull/1175#issuecomment-317918081)? I've created an Admin blueprint for `new_post.yml` as per the Grav Cookbook documentation but, as expected, when my page is getting created via the `+ Add New Post` button in the bar I only have the `title`, `date` and `author` frontmatter set. I want to pass a new frontmatter for `hidemenu: true` but I can't get that passed to the frontmatter on page creation.

In my `new_post.yml` I added this:

```yaml
data:
    hidemenu:
        type:    hidden
        default: 'true'
```

I assumed that if I used a `data:` array that is would trigger `'data' => $data` and pass the `hidemenu: true` to the frontmatter on page creation but it's not.

I know I'm missing something really basic but after about 5 hours of trying different things, I simply can't get this to work.

Any assistance would be fantastic.

PR #1183, opened 2017-07-29
Search pages by slug and folder name too

masetto:
In search bar to search pages not only by title and content, but by slug and folder name too.


rhukster:
Thanks.

PR #1195, opened 2017-08-14
Fix https -> http redirect issue

### LINKED ISSUE #1139: In reverse proxy configurations with https, login attempts to redirect to http
Hey there, we have a reverse proxy configuration where Grav itself is running over http port 8080 behind a Kong proxy. Kong handles serving over https and the ssl termination. 

With this setup, login itself works fine but it is followed up with a redirect to /admin over http, which fails to load in this setup. If I attempt to go to https again, I am already logged in so I know this is a redirect server-side after getting through.

My initial reaction is that the login.html.twig file should use `uri.url(false)` instead of `uri.url` so exclude the hostname that Grav thinks is correct. It's either that or Grav should factor reverse proxy configuration into `uri.url` values. I'm new to Grav so I'm unsure where the correct fix is but modifying the login twig file does seem to fix my issue. If that is acceptable, I can submit the PR.

herrbischoff:
This fixes #1139.

### DIFF
--- a/themes/grav/templates/partials/login.html.twig
+++ b/themes/grav/templates/partials/login.html.twig
@@ -16,7 +16,7 @@
         <form method="post" action="{{ base_url_relative }}">
             <div class="padding">
                 {% block form %}{% endblock %}
-                <input type="hidden" name="redirect" value="{{ uri.url }}" />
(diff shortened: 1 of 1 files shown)


rhukster:
Thanks!

PR #1209, opened 2017-08-28
Update Russian language

DebuggeR94:
Add some missing strings for Russian language.


rhukster:
Please update the language file here: http://translate.getgrav.org/ 

Thanks!

PR #1213, opened 2017-09-04
Fixed language switcher not working on MS Edge

TomOne:
Adapted the `form` attribute polyfill to trigger a submit if any button with a corresponding `form` attribute is clicked. This matches the behaviour of browsers that support the `form` attribute.


TomOne:
Tested and working on MS Edge v14.

rhukster:
Thanks!

PR #1221, opened 2017-09-12
Add support for Latin Extended fonts

newbthenewbd:
The Latin Extended characters, if used, are usually used in conjunction with characters coming purely from the Latin alphabet. This is the case for the Polish language, just to name an example, and most if not all other Slavic languages using the Latin alphabet are likely affected as well. Due to that, I believe that they should be included in the downloads from Google Fonts where possible, since random substitute font appearing in the middle of a word isn't particularly fancy.

I haven't found a guideline containing the information as to whether the compiled CSS should be included, so I haven't compiled it, which seemed like the more sensible choice.

### DIFF
--- a/themes/grav/scss/configuration/fonts/_support.scss
+++ b/themes/grav/scss/configuration/fonts/_support.scss
@@ -35,7 +35,7 @@
         }
     }
 
-    @return $url;
+    @return $url + "&subset=latin-ext";
 }
 
 @mixin body-fonts($font) {


rhukster:
Can you tell me a of a language (polish) and place i can validate in the admin where this is occurring? I need this to ensure the fix is working prperly.

newbthenewbd:
Polish language (selectable in the admin panel as "Polski"), admin panel's dashboard. Very visible in the strings "Wyczyść pamięć podręczną" (Clear Cache) and "Sprawdź aktualizacje" (Check for Updates). These buttons use the font Lato, which is the very first one in the download's URL, and the fix, placed at the end of the URL, makes the text render properly, so I suppose it works fine. Upon closer look, one can see that the problem also occurs in many other places where characters unique to the Polish alphabet are used. Screenshots attached.
Wrong
Right

PR #1222, opened 2017-09-13
Fix [required] fields not working in Microsoft Edge with Selectize.js

### LINKED ISSUE #1061: Adding new page in Edge
The button to create a new page does not work I'm edge. Tested on windows phone and desktop.

TomOne:
This is achieved by adding a Selectize plugin that overrides the bugged `refreshValidityState` method in the Selectize library. This commit should be reverted if selectize/selectize.js/pull/1320 is ever merged and a new version of Selectize gets released.

MS Edge is sometimes the only browser people can use in corporate environments. Even if supporting MS Edge is sometimes cumbersome it would be great if the Grav Admin Plugin would work with Edge without problems. This PR is a step towards it.

I’m aware of the fact that overriding things in external libraries is generally a bad idea. Something could break when the library is updated. But I think this is a special case because Selectize has been looking for new maintainers since 2015 (see selectize/selectize.js/issues/752) and receives only minimal attention from the original maintainer since then. Therefore it’s unlikely that this issue is fixed in Selectize itself (selectize/selectize.js/pull/1320), leaving the Admin plugin forever with this issue.

I tested the changes from this PR in Chrome and in MS Edge without problems.

Fixes #1061

### DIFF
--- a/themes/grav/app/forms/fields/filepicker.js
+++ b/themes/grav/app/forms/fields/filepicker.js
@@ -96,6 +96,7 @@ export default class FilePickerField {
         };
 
         field.selectize({
+            plugins: ['required-fix'],
             valueField: 'name',
             labelField: 'name',
             searchField: 'name',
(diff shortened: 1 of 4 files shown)


rhukster:
merged @w00fz

w00fz:
Cheers, recompiled.

PR #1224, opened 2017-09-14
Fix editor not clickable in list field

### LINKED ISSUE #1168: "editor" field is not clickable if it is in "list" field
If I use `editor` field for child item of `list` field, `editor` is not clickable after clicking `Add Item` button.

To be able to focus on editor and enter text, I need to click on an icon in the toolbar, eg Bold icon, to insert something to editor.

I've also tried to add `autofocus: true` attribute but it only brings cursor into editor, I still can't input anything.

Here is a sample blueprint to test with Antimatter theme, you can save it as user/themes/antimatter/blueprints/default.yaml, edit a page which uses `default` template in admin to see the problem.

A similar problem if we use `codemirror`.

Grav v1.3.1
Admin plugin v1.5.1

```
title: Default
'@extends':
    type: default

title: Default

form:
  fields:
    tabs:
      type: tabs
      active: 1
      fields:
        test:
          type: tab
          title: Test
          fields:
            header.items:
              type: list
              label: Items
              fields:
                .item:
                  type: editor
                  label: Item
```

TomOne:
Editors in lists are not focusable by a click in Chrome, MS Edge an probably some other browsers. Removing the `filter` property from the sortablejs instance resolves this issue and seems to have no other impact on the functionality.

Even without `filter: '.CodeMirror, .grav-editor-resizer'`, dragging inside the editor and dragging the editor resizer sill doesn’t trigger sortablejs, which is desired.

Successfully tested on Chrome 61, Firefox 55, Edge 15 and IE 11.

Fixes #1168

### DIFF
--- a/themes/grav/app/forms/fields/collections.js
+++ b/themes/grav/app/forms/fields/collections.js
@@ -32,7 +32,6 @@ export default class CollectionsField {
                 forceFallback: false,
                 handle: '.collection-sort',
                 animation: 150,
-                filter: '.CodeMirror, .grav-editor-resizer',
                 onUpdate: () => this.reindex(container)
             }));
         });


rhukster:
Thanks, merged and JS rebuilt for next release.

PR #1231, opened 2017-09-23
Add collapsible attribute to lists

newbthenewbd:
It doesn't occur to me that collapsing a list with a single field in it has any use aside from looking ugly. Fieldsets already have this, too, so here we go. The attribute defaults to true, so both backward and forward compatibility should be preserved just fine. (well, if an older grav-plugin-admin version tried to load a blueprint with this attribute set to 0, it would just be ignored, I suppose that's nice enough compatibility)

### DIFF
--- a/themes/grav/templates/forms/fields/list/list.html.twig
+++ b/themes/grav/templates/forms/fields/list/list.html.twig
@@ -65,7 +65,7 @@
                 {% if field.fields %}
                 {% for key, val in value %}
                     {% set itemName = name ? name ~ '.' ~ key : key %}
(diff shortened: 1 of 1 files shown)


w00fz:
This is not going to work, the check should be `field.collapsible is not defined or field.collapsible`.
I updated it for you! I like the idea and consistency with fieldsets, thanks!

newbthenewbd:
Thank You for the update and merging!

In my tests, `is empty` worked, but I sure do not have enough expertise with Twig to rate as to which way is the correct one. After the update, however, there is still a single `is empty` check at line 68 of list.html.twig. Is that intentional?

PR #1232, opened 2017-09-26
Replace a leftover "is empty" check of list's collapsible attribute

newbthenewbd:
Before #1231 was merged, @w00fz mentioned that `is empty` wouldn't work with undefined field.collapsible, and updated the commit. One of the four `is empty` checks has however apparently been omitted, so there's the pull request attempting to correct that.


rhukster:
Thanks missed that one too when reviewing.

w00fz:
Oops, Nice catch!

Thanks!

Issue #1229, opened 2017-09-21
Unable to clear content of select fields (multiple:true) in plugin configuration forms


parcand:
![capture](https://user-images.githubusercontent.com/22176950/30705803-811037c2-9ec4-11e7-9d61-78fb1495e6fb.gif)

This problem does not occur when saving pages data/frontmatter. It only seems to affect config files. I suspect that it could have something to do with the way arrays are being merged in config files. 

```yaml       
someselect:
  type: select
  label: Preface
  multiple: true
  options:
    'ABC': 'ABC'
    'DEF': 'DEF'
    'GHI': 'GHI'
  use: keys
  validate:
    type: commalist
```
[someplugin.zip](https://github.com/getgrav/grav-plugin-admin/files/1322109/someplugin.zip)

parcand:
The problem is quite obvious when using the `toggleable:true` attribute.

![capture2](https://user-images.githubusercontent.com/22176950/30712658-c805d360-9eda-11e7-8d0c-87d85057076d.gif)

rhukster:
I know this was fixed a while back but it seem to have reappeared.  I'll check it out.

rhukster:
I think the problem is that this field is actually a select box underneath with the ability to choose multiple items.  What that means is that select boxes only work when you select 'something'.  If you remove all the items, the select is sent without any data, so its as if it's not sent at all, so the old value is used.

Really the only way I can think to get around this is to create a 'None' Item that you must pick from if you don't want any:

```
      selectize:
        allowEmptyOption: true
      options:
        '': 'None'
        'ABC': 'ABC'
        'DEF': 'DEF'
        'GHI': 'GHI'
```

The allowEmptyOption is required or the default behavior is to use the empty value as the `placeholder`.

rhukster:
I actually have a better option:

```
    someselect:
      type: selectize
      label: Preface
      selectize:
        options:
          - text: ABC
            value: ABC
          - text: DEF
            value: DEF
          - text: GHI
            value: GHI
      validate:
        type: commalist
```

This uses the selectize comma list and passes the options.  More information here: https://github.com/selectize/selectize.js/blob/master/docs/usage.md

parcand:
That's a pretty good solution. There seems to be one little glitch though. Selected items are properly saved but for some reason, using the `options` setting prevents the selectize widget from properly displaying more than one selected item when the page is reloaded after being saved.

[...]

PR #1236, opened 2017-10-02
Fixed selecitze field not properly updating its value when the `options` attribute is provided and more than one option is selected.

parcand:
See #1229

### DIFF
--- a/themes/grav/templates/forms/fields/selectize/selectize.html.twig
+++ b/themes/grav/templates/forms/fields/selectize/selectize.html.twig
@@ -1,7 +1,7 @@
 {% extends "forms/field.html.twig" %}
 
 {% block global_attributes %}
-    data-grav-selectize="{{ (field.selectize is defined ? field.selectize|merge({'create': true}) : {'create': true})|json_encode()|e('html_attr') }}"
+    data-grav-selectize="{{ (field.selectize is defined ? field.selectize|merge({'create': true})|merge({'items':value}) : {'create': true})|json_encode()|e('html_attr') }}"
     {{ parent() }}
 {% endblock %}
 


rhukster:
I'm not really able to get this to work, even with your last field definition:

```
  someselect:
      type: selectize
      label: Preface
      selectize:
        items:
          - ABC
          - DEF
        options:
          - text: ABC
            value: ABC
          - text: DEF
            value: DEF
          - text: GHI
            value: GHI
      validate:
        type: commalist
```

Is this what you have? What's the purpose of the `selectize: items:` ? No matter what i put i don't get any values saved.

parcand:
Oh, I'm sorry, I can see where the confusion comes from.

I am no longer providing the `selectize: items:` as part of the field definition. This pull request introduces a small change which now automatically populate the`selectize: items:` from the actual field value.

Just to give you a bit of background information, it appears that using the `selectize: options:` attribute leads to a small glitch which prevents the widget from properly loading the comma-separated values from the HTML input element  itself. For some reason it only loads a single item, even if you previously saved more that one.

So in order to fix that problem you need to pass pre-selected values directly to the widget using the`selectize: items:`. It would be quite impractical to do it at the blueprint level, that's why I came up with this solution which merge values directly in the twig template of the selectize widget.

```yaml
    someselectize:
      type: selectize
      label: A selectize field
      selectize:
        options:
          - text: ABC
            value: ABC
          - text: DEF
            value: DEF
          - text: GHI
            value: GHI
      validate:
        type: commalist
```

parcand:
I've done some tests with an up-to-date Grav installation and it seems to work as expected, for both config blueprints and pages blueprints (with the `header` field name prefix).

gvstavnev:
is there a way to pass data with `@data-options` to `selectize: options:` attribute?

PR #1239, opened 2017-10-05
Fix Dropzone upload timeout

### LINKED ISSUE #1238: Unable to upload large files on slow internet connections
Currently its not possible to upload large files (e.g. PDFs) on slow internet connections.
When the upload takes longer then 30 seconds, it just aborts.

This is due the default `timeout` option set by [Dropzone](http://www.dropzonejs.com/#config-timeout).
To work around it you can just set the `timeout` option to `0` which disables the timeout completely.

jackmu95:
Closes #1238.

### DIFF
--- a/themes/grav/app/forms/fields/files.js
+++ b/themes/grav/app/forms/fields/files.js
@@ -46,7 +46,7 @@ Dropzone.confirm = (question, accepted, rejected) => {
 };
 
 const DropzoneMediaConfig = {
-    createImageThumbnails: { },
+    timeout: 0,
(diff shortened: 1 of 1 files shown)


pieplu:
I apply this patch on one of my website and it fix the bug. Thanks a lot @jackmu95 !
(Grav: **1.3.3** - admin **1.6.0**)

rhukster:
Thanks merged.. Will ensure we rebuild JS before next release

w00fz:
I recompiled and added the CHANGELOG entry 👍

PR #1248, opened 2017-10-19
Add/Change some terms translations (French locale)

kernox:


rhukster:
He thanks for this but all translations for the admin are handled through crowdin. 

https://getgrav.org/blog/crowdin-translation

PR #1267, opened 2017-11-11
Added code to use GPM inline with grav PR #1746

aguvillalba:
So now it calls the new method GPM::loadRemoteGrav() when creating the static GPM

### DIFF
--- a/classes/gpm.php
+++ b/classes/gpm.php
@@ -27,7 +27,9 @@ class Gpm
     {
         if (!static::$GPM) {
             static::$GPM = new GravGPM();
-            static::$GPM->loadRemoteGrav();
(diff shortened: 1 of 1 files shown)


classes/gpm.php
```diff
@@ -27,6 +27,7 @@ public static function GPM()
     {
         if (!static::$GPM) {
             static::$GPM = new GravGPM();
+            static::$GPM->loadRemoteGrav();
```
rhukster: As a quick way to make this 'safer' and backward compatible, this could be to wrap it in a `method_exists` check.  Something like:

```
if (method_exists('GravGPM', 'loadRemoteGrav') {
  static::$GPM->loadRemoteGrav();
} else {
  static::$GPM = new GravGPM();
}
```

Issue #1035, opened 2017-03-23
Broken indentation when selecting parent page


phideas:
Using long Titles in pages breaks the indentation in list when selecting parent page
making it hard for user to search for parent page.
http://imgur.com/a/8GLan

Suggestion: indent list items with `padding-left` css instead multiple `&nbsp;`

phideas:
Other related issue is too much potential parent pages in list. I can have a blog or web with 100s of pages but wanna see only 12 relevant pages (first 2 levels) as possible parent pages. It would make the list much more clear and easy to select parent. 

Suggestion: implement an option to limit parent pages in list by level depth.

Issue #941, opened 2017-01-20
Can't scroll in "Parent Page" select when too much pages are created 


saxinte:
Hello Grav Team :)

I just found that if you have too many pages created in your Grav App (I have more than 200 pages), then you can't found them within the "Add page" feature in the Admin panel.

Tested with Grav 1.0.10 and Grav Admin 1.0.10: bug is present
Tested with Grav 1.1.13 and Grav Admin 1.2.8: bug is present

How to replicate:
- Create a subfolder within your app in `/user/pages/a` http://imgur.com/tMe3Xyb
- Add 200 pages in it.
- Open the admin, then click on "Pages" on the Left Hand Panel.
- Click "Add Page" at the top-right.
- In the "Parent page" select, try to access a page that is not inside this `a` folder. It's impossible.

I used `a` folder in order to have all those pages displayed at the top of the "Parent page" select. Pages above can't be reached.

I can find my page in typing it directly but it's hard to find it. It's also very laggy because of the amount of DOM node here. A better solution would be either a real `<select>` element or to be able to look for a given path, not the page title.

I've added a screenshot and a video to show you the issue (In the video I show you that I only see the pages within the `/a`folder - I can't access other pages)

- Pages: http://imgur.com/tMe3Xyb
- Video: http://imgur.com/QxLKSxX

Thanks,
Sebastien

phideas:
**Feature request:**
I can confirm this issue is a real problem. As the number of website pages grows, it is getting hard to select parent when trying to add new page. I think a new aproach to select parent is needed here. I do not know the right answer to this problem, but something that would show only the requiered top levels is needed.

rhukster:
What browser is this?  I'm trying the parent dropdown with the latest release and it's fine even with 1000+ pages.

phideas:
Sorry for confusion. Can't recreate this bug either. My complain is related to UX when a lot of pages in list makes it hard for user to quickly select parent. Gonna create a feature request.

saxinte:
Hello guys, sorry being away for long, I was busy working on other stuff.
I just updated to Grav 1.2.2 and Admin 1.3.3 and even if the way Grav displays the pages and child, the issue is still their.

I've recorded a video to show you the issue. Basically, I can see my page from Alphabetic Order with the parent page and the modules within. But when I reach the end of the scroll I don't see the remaining pages.

The search page works if I'm start typing a page which is fine, but not the scroll.
[...]

Issue #906, opened 2016-12-29
New Page Dialogue Slow to Select Parent


trst:
I have a blog with a few hundred posts (and growing) set as children to the home page. When creating a new Post, the drop down for `Parent` takes ages to load and has an adverse affect on the UI of the form. Performance is fine until the drop down is active.

I can think of a few work arounds:

1. Limit how many pages are available before searching or using a plain text input
2. Providing some mechanism to designate pages that can be parents (explicitly in frontmatter, perhaps?)
3. Allowing some way of creating a page directly under a specified parent (perhaps as an interface plugin to add another button)

Thoughts?

flaviocopes:
I tested with 400+ pages and beside a slower loading of Admin (1.5s vs 0.7 with a few pages), to me the select opens quite fast (maybe 200ms of delay). On Firefox. Can you make a small screencast? Which browser are you using?

Anyway determining which pages can be parents, is a good idea to me, for example for a blog I'd never (or almost never) use a blog post as a parent. Maybe limited to some page types (e.g. I choose a blog post (`Item`), the parent must be a `Blog` page.

trst:
I've been testing from both Safari on macOS (El Capitan) and Mobile Safari
(iPhone 6+ and iPad Pro 12.9"). I'll check in other browsers shortly,
perhaps I'll check to make sure I'm not experiencing some strange caching
issue.

Issue #902, opened 2016-12-24
Feature request: a way to constrain the options for the parent page in the new page dialogue


SarahBaghdadi:
- So that users can't accidentally put pages where they don't belong.
- To simplify the new page dialogue.

PR #1298, opened 2017-12-24
Limit Parents Levels

hugoaf:
The main beneffit of this PR is to improve the parent selection when you have many pages.

This PR will add a field in blueprint and a limit var to the `parents` field in order to limit the levels of parents to show in admin forms. The parent levels limit filter was already implemented in `pages` field, so there was not much to change to make it work.

Solves part or act as a workarround for these issues:
#735, #902, #906,  #941, #1035 

In the future, this might be improved adding low and high level limits. Also having parents filter by page type would be great, or even adding a header `parent: false` by page basis. But for now this PR helps.

### DIFF
--- a/blueprints.yaml
+++ b/blueprints.yaml
@@ -157,6 +157,12 @@ form:
         folder: Show folder
         fullpath: Show fullpath
 
+    pages.parents_levels:
+      type: text
+      label: Parents Levels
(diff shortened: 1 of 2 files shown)


rhukster:
Thanks, will be in next release

Issue #1024, opened 2017-03-16
"Array to string conversion" exception when saving a page with field type "pages" and "multiple: true"


benface:
Tested on new installs of
- Grav v1.1.17 - Admin v1.2.14
- Grav v1.2.0-rc.1 - Admin v1.3.0-rc.1

Environment:
- macOS 10.12.3
- Safari 10.0.3
- localhost (with MAMP PRO)
- PHP 7.1

Repro steps:
1. Install Grav with the admin plugin.
2. Open `system/blueprints/pages/default.yaml` (just for speed, but I can reproduce the bug with a page blueprint found in a theme as well)
3. In the `content` tab, add this field (line 38):
```
            header.content.items:
              label: Pages
              type: pages
              multiple: true
```
4. Edit the Home page through the admin panel.
5. In the "Content" tab, scroll to the bottom and enter "Home" in the "Pages" field. Select the Home page from the dropdown so it becomes a token in the input.
6. Try to save the page.

Expected result: No error.
Actual result:
```
Whoops \ Exception \ ErrorException (E_NOTICE)
Array to string conversion
```

This does not happen if the field doesn't have `multiple: true`. I also tried adding `validation: type: array` but it doesn't fix it.

benface:
Update: This error doesn't occur when
```
validation:
  type: commalist
```
is added to the field in the blueprint.

rhukster:
FYI, you shouldn't be modifying the system blueprints.  If you want to add a field you should do it with an override. Look at the antimatter theme in the `blueprints/pages` folder for some relevant examples.

benface:
I know, this happens when doing it in a theme as well. So this is not considered a bug? I think at least the error message could be a little more helpful.

rhukster:
hmm.. I guess we could auto-handle forcing comma list in pages field when multiple is true.. I'll open her up again.

baublet:
Can confirm that this is a bug, and that adding:

```
validate:  # NOT `validation`
  type: commalist
```

to the field in question **does** seem to work for my case. The field in its entirety being:

```
            header.associated_products:
              type: pages
              multiple: true
              label: Products applicable for this warranty
              validate:
                type: commalist
```

To reproduce the exception OP was talking about, just remove the two lines beginning at `validate` and try to save the page.

Amorfx:
I have the same problem with filepicker or pagemediaselect when saving i've got this error.
`
header.brand.img_nav.:
              label: Image nav
[...]

PR #1364, opened 2018-03-03
Add support for 'default' and 'options' for taxonomy field

pshe94:
With this patch user can override in blueprint the taxonomy field to add custom taxonomy values in `default` and `options` attributes.
There is only one catch - when overriding default taxonomy field, with both `default` and `options` attributes, you need to add `validation` attribute with `type: commalist`, as shown in example below. Otherwise, after first save Grav will throw an error at top of the page `Array to string conversion`, and page won't be saved. But after second try, it will be successful. I haven't found why this happens, but adding validation fixes it. I'm guessing that it might be related to #1024.

```
taxonomies:
  fields:
    header.taxonomy:
      default:
        category: ['blog','page']
        tag: ['test']
      options:
        category: ['grav']
      validate:
        type: commalist
```


rhukster:
Nice improvement 👍

Genenenenaam:
Just what i was looking for GREAT :)

Do add this info to the documentation please

PR #1405, opened 2018-03-30
Proposed new Event onAdminLogin

ricardo118:
this PR proposed a new event `onAdminLogin` this can be particularly useful for plugins that require a one time processing Event after the user has been authenticated successfully.

### DIFF
--- a/classes/admin.php
+++ b/classes/admin.php
@@ -410,7 +410,8 @@ class Admin
 
             unset($this->grav['user']);
             $this->grav['user'] = $user;
-
+            
+            $this->grav->fireEvent('onAdminLogin');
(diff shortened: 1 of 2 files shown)


rhukster:
I like this idea, but a `UserLoginEvent` object should be passed like the `onUserLogin` event in login plugin

```
            $event = new UserLoginEvent($event->toArray());
            $grav->fireEvent('onUserLogin', $event);
```

This will give access to data that could prove very useful for any plugin wanting to take advantage of the `onAdminLogin` event.

ricardo118:
i tried adding what you mentioned, and importing the userloginevent but i still dont know what do and the error on `$event->toArray()`

rhukster:
Admin now uses standard login events.

Issue #1408, opened 2018-04-01
Error while trying to copy a page


underr-ua:
Admin 1.7.3

Error while trying to copy a page: 0 - Call to a member function blueprints() on null

PR #1409, opened 2018-04-02
Update admin.php

drnasin:
Fix for https://github.com/getgrav/grav-plugin-admin/issues/1408


rhukster:
Darn I hadn’t checked a page copy! Thanks.

PR #1410, opened 2018-04-02
Fix collapsed noncollapsible list fields

newbthenewbd:
The value of `field.collapsed` seems to be true even in the case of collapsible fields, making noncollapsible fields collapsed by default, with no way to expand them. According to a quick test I've done, this should fix that.


newbthenewbd:
Actually, there's a deeper issue with this, as the collapsible fields now randomly appear expanded, yet their buttons are set the other way around... At least this serves as a somewhat better issue trace than the Slack chat log :upside_down_face:

newbthenewbd:
Here we go, should be good now.

PR #1415, opened 2018-04-05
Update blueprints.yaml

andrewvaca:
Missing "step" attribute for quality causing Safari browser (I didn't test it on other browsers) uses default step "1" and invalidate the quality input due the default input value is 0.8. This is preventing whole form from submit. Adding the step attribute fixes the bug.


mahagr:
Looks like valid fix, cc @rhukster

PR #1422, opened 2018-04-12
Small addition to make rows stand out better on hover

rexfordkelly:
Added a darker background to rows on hover, to ensure they stand out better. Added a new "$row-hover" variable and updated a few styles to use it.


rhukster:
Why not just change the lines:

```
background: lighten($content-bg, 2%);
```

to:

```
background: lighten($content-bg, 4%);
```

Adding a new variable will break some plugins i have that create different presets

PR #1429, opened 2018-04-19
Debug resolution tests during image file upload

guins:
Hi, 
I came across the bug where resolution tests where always true.
Here's what I changed :

1/ I added `resolution: settings.resolution` to the Field options so the `resolution` is actually available within the `accept` callback (`ACCEPT_FUNC`). Because the `accept` callback is always overridden in the Field constructor and the `resolution` option wasn't passed down here.

2/ I made sure the resolution tests are executed :
- Made sure it's an image as the resolution tests only make sens for images
- Wait for the image to be loaded before testing the width & height.

I guess the bug is also available in the form plugin.
Hope it helps !

### DIFF
--- a/themes/grav/app/forms/fields/files.js
+++ b/themes/grav/app/forms/fields/files.js
@@ -72,13 +72,17 @@ const DropzoneMediaConfig = {
 
 const ACCEPT_FUNC = function(file, done, settings) {
     const resolution = settings.resolution;
-    if (!resolution) return done();
+    const isImage = file['type'] && !!file['type'].match(/^image\//);
+    if (!isImage || !resolution) return done();
 
-    setTimeout(() => {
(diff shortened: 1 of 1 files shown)


w00fz:
@guins thanks, good catch on the resolution setting. I guess I forgot to pass that along!

I'm not entirely sure about creating a new `Image` object and loading the source again. All the image metadata should be already available in the `file` object (sizes/mime type/etc), taken care by dropzone when you drop in/browser a file. 

I guess it is definitely better than the 50ms timeout from before but I wish there was a better way as now we are previewing the image twice and on big sized images it could add some stress to the browser.

w00fz:
This was implemented very similarly to yours, I just added a few nice additions from your PR.

Thanks!

PR #1445, opened 2018-05-14
Fix for plugin name mismatch

jgonyea:
Replaces suggested plugin name with the corrected name.


rhukster:
Thanks, kinda had forgotten!

PR #1451, opened 2018-05-26
Initialize session on setup

tribut:
This makes the plugin compatible with `system.session.initialize = false` which makes it possible to have no session cookie (for frontend users) while the admin login still works.

The change seems to work fine, but I'm pretty new to Grav and it is entirely possible that the initialization should happen somewhere in the login or form plugins. Any feedback appreciated.

### DIFF
--- a/admin.php
+++ b/admin.php
@@ -144,6 +144,7 @@ class AdminPlugin extends Plugin
 
         // Only activate admin if we're inside the admin path.
         if ($this->isAdminPath()) {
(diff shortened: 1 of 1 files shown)


mahagr:
I think the fix itself is fine, though as session used to be opened very early (before plugins were initialized), some plugins may break if the `system.session.initialize = false` was used, which is why it is disabled as default.

You're firing the event very early, so changes of breaking anything (except for non-admin plugins) is very low.

PR #1463, opened 2018-06-07
Fix for FILEPICKER fields that are a child field of a list

hacktisch:
I found that when a Filepicker field has a custom source folder defined, that didn't work when the filepicker was a child field (of a list). The blueprint settings wouldn't be found by this method. My proposed fix is what resolved the issue for me but I am not sure that this is 100% sound and conforms to the Grav paradigm. I'm hoping a correct fix will be included in the next update because I have no idea how I could hook this functionality properly within Grav (I'm new to Grav)

### DIFF
--- a/classes/adminbasecontroller.php
+++ b/classes/adminbasecontroller.php
@@ -751,7 +751,7 @@ class AdminBaseController
             return false;
         }
 
-        if ($data instanceof Data) {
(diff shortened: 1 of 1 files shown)


mahagr:
Looks like this has already been dealt with and released.

PR #1480, opened 2018-07-06
Change usage of basename where possible

tcyrus:
`DirectoryIterator::getFilename` returns the same value as `DirectoryIterator::getBasename` (no argument) while being locale safe (See getgrav/grav#2087). `DirectoryIterator::getBasename` calls `basename` using the argument provided and returns the result, so calling `basename` on the result can be replaced with passing the argument to `DirectoryIterator::getBasename`. I replaced one call to `getBasename` with `getFilename` as well as simplifying `basename` calls on `DirectoryIterator::getBasename`. I also simplified some usage of `pathinfo`, another locale aware function.

There is still an issue with locale safety (`basename` is still called by `DirectoryIterator::getBasename`). I have mentioned some possible solutions in my issue to the main Grav repository (getgrav/grav#2084).

### DIFF
--- a/classes/admin.php
+++ b/classes/admin.php
@@ -669,10 +669,10 @@ class Admin
             $obj->file = $file;
             $obj->page = $this->grav['pages']->get(dirname($obj->path));
 
-            $filename = pathinfo($obj->title)['filename'];
-            $filename = str_replace(['@3x', '@2x'], '', $filename);
-            if (isset(pathinfo($obj->title)['extension'])) {
-                $filename .= '.' . pathinfo($obj->title)['extension'];
+            $fileInfo = pathinfo($obj->title);
+            $filename = str_replace(['@3x', '@2x'], '', $fileInfo['filename']);
+            if (isset($fileInfo['extension'])) {
+                $filename .= '.' . $fileInfo['extension'];
             }
 
             if ($obj->page && isset($obj->page->media()[$filename])) {
--- a/classes/admin.php
+++ b/classes/admin.php
@@ -170,11 +170,11 @@ class Admin
 
         /** @var \DirectoryIterator $directory */
         foreach (new \DirectoryIterator($path) as $file) {
-            if ($file->isDir() || $file->isDot() || Utils::startsWith($file->getBasename(), '.')) {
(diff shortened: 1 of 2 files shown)


classes/admin.php
```diff
@@ -170,11 +170,11 @@ public static function adminLanguages()
 
         /** @var \DirectoryIterator $directory */
         foreach (new \DirectoryIterator($path) as $file) {
-            if ($file->isDir() || $file->isDot() || Utils::startsWith($file->getBasename(), '.')) {
+            if ($file->isDir() || $file->isDot() || Utils::startsWith($file->getFilename(), '.')) {
                 continue;
             }
 
-            $lang = basename($file->getBasename(), '.yaml');
+            $lang = $file->getBasename('.yaml');
```
mahagr: Didn't you say that the method is not locale friendly? How is this better than `basename($file->getFilename(), '.yaml');`?

tcyrus: The reason that `DirectoryIterator::getBasename` and `SplFileInfo::getBasename` are not locale friendly is because they call `basename` underneath (See [implementation](https://github.com/php/php-src/blob/master/ext/spl/spl_directory.c#L997-L1041)). In other words

```php
(basename($file->getBasename(), '.yaml') === $file->getBasename('.yaml')) &&
($file->getBasename('.yaml') === basename($file->getFilename(), '.yaml'))
```

tcyrus: It also matches up with the usage of `DirectoryIterator::getBasename` in the main grav repository (See [here](https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Data/Blueprints.php#L66-L73) and [here](https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Themes.php#L302-L308))

tcyrus: @mahagr, does this answer your question?

mahagr:
Looks good to me!

tcyrus:
@mahagr, whats the status of this PR?

mahagr:
I am not sure. There is no second review, so maybe it was just accidentally missed up. I am quite busy right now, if nothing happens, please remind me in a week, ok?

tcyrus:
@mahagr, Just a reminder

PR #1491, opened 2018-07-25
Add the form attribute polyfill again

TomOne:
The _themes/grav/js/form-attr.polyfill.js_ file was deleted in c37988f68104e83ca7149d222a37269567f400d2. Was this intentional? The commit message (“CSS rebuild again”) seems unrelated to it.

The removal of the polyfill led to the problem again that saving in the admin panel doesn’t work in IE 11 / Edge.

### DIFF
--- a/themes/grav/js/form-attr.polyfill.js
+++ b/themes/grav/js/form-attr.polyfill.js
@@ -1,0 +1,80 @@
+(function($) {
+    $(function() {
+        /**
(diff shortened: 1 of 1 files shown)


TomOne:
Could this PR be merged? Without it the Admin Panel is unusable in Internet Explorer / Edge.

rhukster:
Not intentional, not sure why that was removed. Thanks for adding back.

PR #1533, opened 2018-10-22
Hotfix for admincontroller.php

zouloux:
Now checking if 'content' is in $data. Some blueprints can have 'content' optional so not defined, which cause a crash when editing.
Checking with operator `??` and default value empty string `''`.


rhukster:
Had to revert due to php7 syntax.

zouloux:
Oops ! My bad. Do you need a new PR ?
Thanks

rhukster:
nah i'll just fix it

PR #1557, opened 2018-12-13
Admin in site root on subdomain

igk1972:
Fixes for case:
- admin base '/'.
- subdomain (admin.example.com).
- grav core (without admin) work on main domain.


themes/grav/templates/partials/logo.html.twig
```diff
@@ -1,20 +1,16 @@
 {% if config.plugins.admin.logo_text %}
 <h3>
-	<a href="{{ base_url_relative }}">
+	<a href="{{ base_url_relative }}/">
```
mahagr: We really need to do the fix without adding slashes to every `base_url_relative` usage. This is because of your change will break all existing plugins which are relying on the current behaviour.

PR #1559, opened 2018-12-14
Fix unset all images bug

hdwebpros:
This should at least fix the bug where unset was changing all dropzone images within the admin

### DIFF
--- a/themes/grav/app/pages/page/unset.js
+++ b/themes/grav/app/pages/page/unset.js
@@ -1,15 +1,19 @@
 import $ from 'jquery';
 
 $(document).on('click', '.dz-unset', function() {
-    $(this).closest('.dz-image-preview').remove();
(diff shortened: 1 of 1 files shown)


hdwebpros:
I tried. I created a branch called `fix-unset-all-images-bug`.

For some reason, that didn't take. I'll try again I guess

hdwebpros:
I tried again. I can't figure it out. My old commits show up no matter what. Even with a different branch. Can you please help and perhaps commit that code for me? I'm not sure what else I can do

w00fz:
Unfortunately this is not ok. You are missing the bigger picture which is 1.9beta

A lot has gone towards the media including under so probably most issues you are seeing and trying to fix are already sorted.

https://github.com/getgrav/grav-plugin-admin/blob/1.9/themes/grav/app/pages/page/unset.js

PR #1615, opened 2019-02-12
Update parents.html.twig

drnasin:
This commit fixes the issue - https://github.com/getgrav/grav/issues/2362


rhukster:
This actually changes behavior, i've already fixed with a simpler solution.

PR #1631, opened 2019-03-13
Non-checked toggle-highlight

OleVik:
Change toggle-field highlight to white on grey, checked to white on blue. This allows using `highlight`-property on toggle-field without `default`-property.

Highlighted value:
image

Checked value:
image

Suggested for improving UX, where defaults can be highlighted without having to be checked - and thus automatically saved with the Page/file.


rhukster:
Ole, just for clarification what's the problem with current implementation?  right now the whole intention of the 'blue' is to show the default value, and the grey to show selected but not-default.  Changing this behavior is going to have impacts.

OleVik:
Specifically, that using `highlight: 'value` does not have an effect, because it has to be paired with `default: 'value'` for the CSS to take effect. Whilst defaulting to a value isn't bad, saving a Page or other setting enforces the value in the resulting file. Thus I cannot simply highlight the default set by the system or extension, I have to force it into the saved object for it to take effect in CSS.

themes/grav/scss/preset.scss
```diff
@@ -1080,14 +1080,14 @@ form {
         transition: background-color 0.5s ease;
     }
 
-    &.switch-toggle input.highlight:checked + label {
```
OleVik: I may have confused the two; white on blue should then be from highlighting, and white on grey from checked? The main problem is this, where `input.highlight:checked + label` needs to operate without `:checked`.

rhukster: its a radio tho, not a checkbox.. something is always supposed to be checked.

OleVik: If the input-field is required, yes, but that's not necessarily the case. Especially in blueprints, if I default one value to be checked, that for example has the side-effect of being saved to the FrontMatter of a Page. This inadvertently makes explicit what is already implicit from the extension's options, and so to avoid polluting the resulting file I would want to highlight the default rather than check it.

Issue #1427, opened 2018-04-18
Unset Media doesn't work


guins:
Hi, it appears that the "unset" media button (the cross icon when you hover over a preview file) doesn't seem to work.

Setup :
- Grav install: `Grav v1.4.3` & `Grav Admin Panel v1.7.4`
- Have a page with several image file fields with `multiple` attribute set to `false`
- Add a image for each file field
- Save the page
(It should be all good up to here)

My .yaml file :
```.yaml
header.custom.image1:
    type: file
    label: "Image 1"
    filesize: 3
    destination: 'self@'
    accept:
        - image/*

header.custom.image2:
    type: file
    label: "Image 2"
    filesize: 3
    destination: 'self@'
    accept:
        - image/*
```

**Problem 1 :**
- Unset one of the images using the cross icon when you hover the preview
- Notice that the drag & drop zone doesn't display any message now
- Try to upload a new file. You will get the error popin `You can not upload any more files.`

**Problem 2 :**
- Unset one of the images using the cross icon when you hover the preview
- Notice that the dropdown zone doesn't display any message now
- Save the page
Now you should end up with all image file fields, filled with the same image ! (the image is the first filled image field)

@lepascal posted an exemple of this issue [here](https://github.com/getgrav/grav/issues/1982)  with a repeater field, but it's also true with only 2 image fields.

**Temporary Solution 1 - NOT WORKING**
I wanted to set my image file fields as required using the `validate.required` attribute in order to prevent saving the page if an image field is empty, but this is not working. This is a know bug and I added a comment [here](https://github.com/getgrav/grav-plugin-admin/issues/887)

**Temporary Solution 2 - working, but not ideal**
As the problem seems to come from the cross icon only, I got rid of it.
In my theme, I added this piece of code into my `custom.js` file :
```js
$(function(){
    // Hide existing unset buttons
    $('.dz-unset').css('display', 'none');

    // Prevent click on unset buttons
    // In case the button was added dynamically
    // or the display attribute was overridden
    $(document).on('click', '.dz-unset', function (e) {
        e.preventDefault();
        e.stopImmediatePropagation();
        alert('Sorry, the unset button is not working');
    });
});
```

Thanks

lepascal:
> I wanted to set my image file fields as required using the validate.required attribute in order to prevent saving the page if an image field is empty, but this is not working. This is a know bug and I added a comment here

[...]

PR #1670, opened 2019-05-02
Re enable  "unset" button and fix for multiple drop zones

rotanadan:
"unset" on images in dropzones was shut off. I turned it back on and fixed the big issue it had when being used on pages with multiple dropzones.
"unset" will clear the image, but not delete the file from the server. This is requested a lot from clients because I often blueprint things like hero images, backgrounds, etc that are used on multiple pages.


w00fz:
I am confused by this pull request because Unset is not removed as functionality at all.

You have 2 action icons you can pick from, 1 is the ⓧ unset and the other is the 🗑delete. Unset only appears as you drop a new image in the dropzone. Before saving, if you change your mind and want to remove it, you click unset and it removes it from the session. 
Once you save the page, the field with the image is persistently stored and you cannot unset it anymore, now you can only delete it, which is why the unset icon is gone.

unset_vs_delete

(note that when unset is available, also delete is but they perform the same 'unset' action)

This behavior is working fine for me so unless I am misunderstanding the issue, I can't approve this PR.

mmorkt:
It’s easy to recreate the bug he’s fixing. Try placing the dropzone in a list and make some rows. Upload some files into the dropzones and save. Use “unset” on a file and save. Now your page is broken. As mentioned by multiple users https://github.com/getgrav/grav-plugin-admin/issues/1427

hdwebpros:
Please re-open this PR.

@w00fz you might not be understanding what unset is truly for.

I'll give you a practical example.

Let's say you have two pages with the same hero image on each page. Then, you're asked to change one of the images. Can you see where this is going? With the current setup, you can only delete -- which deletes the file from the server. That would delete the image from **_both_** pages. 

We have multiple clients, and ourselves that run into the situation where someone wants to remove an image from a page but *not* delete it from the server. _That_ is where unset comes in. It's not just for when you first place it in a dropzone. It's for removing an image/file/whatever from a page without deleting it from the server.

It's a major issue that we have a few upset clients about. Judging from some of the interactions on Github, we aren't the only ones who see this is a big issue.

This PR addresses the main issue. It should be re-opened and tested.

w00fz:
Also fixes #1675

w00fz:
Thanks for clarifying @mmorkt and @hdwebpros  , it wasn't clear this was about the same image on multiple file fields! Makes sense now and PR is ok.
I cleaned up some merge conflicts and it is now merged.

Cheers all!

PR #1694, opened 2019-06-13
add Dockerfile

setop:
basic docker file which provide a docker image with grav and grav admin


rhukster:
We are working in a standalone repo here: https://github.com/getgrav/docker-grav

PR #1728, opened 2019-07-31
Page media - single file "view" action - use filename instead of extras

### LINKED ISSUE #1727: Page Media single file "View" action doesn't work for most files
The button works for image files (since they have "extras" filled), but doesn't work for most other file types (null/undefined instead of the filename is inserted into the url).

This was introduced in dbdd6e2a45dfee11cc69029f7eae25d825168361 and as far as I can tell there's no reason for the change.

Steps to reproduce:
1) Drop an mp4 or pdf file into page media
2) Save
3) Try using the view button

Result:
Page url with _undefined_ instead of the file name is opened

Expected:
Open the file in a new tab

robertbak:
Fixes #1727

I've tested this with different file types, and all kinds of weird filenames, can't see why the `extras.original` was used (especially since most files don't have that populated).

### DIFF
--- a/themes/grav/app/pages/page/media.js
+++ b/themes/grav/app/pages/page/media.js
@@ -137,11 +137,10 @@ export default class PageMedia extends FilesField {
         this.container.delegate('[data-dz-view]', 'mouseenter', (e) => {
             let target = $(e.currentTarget);
             let file = target.parent('.dz-preview').find('.dz-filename');
-            let filename = encodeURI(file.text());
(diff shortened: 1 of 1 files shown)


rhukster:
@w00fz can you please take a look at this one. Looks ok to me.

mahagr:
Original is used for image resizing.

w00fz:
Thanks for this, you were spot on with the issue, I just fixed it slightly differently.

Cheers

PR #1732, opened 2019-08-07
Fix `accept` for SVG files in `file` uploaders:

arkhi:
Using `accept: 'image/svg+xml' to only allow SVG file to be uploaded should now work.

The `preg_match` was comparing `svgggg…xml` with `svg+xml`.

Check the [List of Media Types](https://www.iana.org/assignments/media-types/media-types.xhtml#image) for more potential issues.


arkhi:
This is a cross PR with getgrav/grav-plugin-form#364.

PR #1752, opened 2019-09-16
Implementation of the setting to enable secure deletion of pages

lauroguedes:
To avoid accidental deletion, I implemented a secure deletion system in which the user has to type the word DELETE into a text field for the confirm deletion button to be enabled. By default this plugin setting is disabled.

demo-secure-delete


rhukster:
This is pretty cool!  Good idea

lauroguedes:
Yes, I got the idea after I accidentally lost some data, as the button was the same color as cancel and there was no turning back this was a solution that solved

lauroguedes:
Another thing that could be done is to have a specific permission to create, edit and delete pages in admin, so that would create another layer of security.

mahagr:
There are already separate permissions for different page actions built into the upcoming Flex pages. :)

PR #1769, opened 2019-10-21
Delete-button UX

### LINKED ISSUE #1768: Use of multiple colour Page Editor button, esp. Red
I know a PR was requested, but I cannot figure out how to make the below change via code.

Recently I've noticed that when editing Pages there are now up to four colors in the top-right Page Editor button row:

While the color choice of red for the "Delete" button is to serve as a warning, it actually seems to draw even more attention to itself as is a distraction and fights for attention with the primary action button in purple.

Perhaps one alternative is to keep the "Delete" button the standard color, but change both the color and label (from the more generic "Continue") of the resulting confirmation dialog:

I look forward to hearing feedback on this proposal.
Paul

OleVik:
Fixes #1768.

### DIFF
--- a/themes/grav/templates/pages.html.twig
+++ b/themes/grav/templates/pages.html.twig
@@ -213,9 +213,9 @@
                 {% endif %}
 
                 {% if warn %}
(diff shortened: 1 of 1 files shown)


paulhibbitts:
Thanks very much @OleVik for putting together this PR.

Please note that the button label for the modal delete dialog "Continue" button with a checkmark is proposed to be changed to "Delete" with a trash can icon (trash-o).

OleVik:
The PR does not change the content of the text, only the classes - reassigning `danger`.

paulhibbitts:
Thanks very much @rhukster and @OleVik.

PR #1780, opened 2019-11-04
Allow create-option to be used

OleVik:
Let the field define the [create](https://github.com/selectize/selectize.js/blob/master/docs/usage.md#general)-option for Selectize.js, defaulting to `true`.

### DIFF
--- a/themes/grav/templates/forms/fields/selectize/selectize.html.twig
+++ b/themes/grav/templates/forms/fields/selectize/selectize.html.twig
@@ -2,7 +2,7 @@
 
 {% block global_attributes %}
     {% if field.selectize is defined %}
-        {% set fieldSelectize = field.selectize|merge({'create': true}) %}
+        {% set fieldSelectize = field.selectize|merge({'create': field.selectize.create ?? true}) %}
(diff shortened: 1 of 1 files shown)


mahagr:
Umh, I would rather just pass selectize variables as they are in YAML with some default, but I can see why we would need that. Maybe turn the merge around?

themes/grav/templates/forms/fields/selectize/selectize.html.twig
```diff
@@ -2,7 +2,7 @@
 
 {% block global_attributes %}
     {% if field.selectize is defined %}
-        {% set fieldSelectize = field.selectize|merge({'create': true}) %}
+        {% set fieldSelectize = field.selectize|merge({'create': field.selectize.create ?? true}) %}
```
mahagr: How about:
```twig
{% set fieldSelectize = {create: field.selectize.create ?? true}|merge(field.selectize ?? {}) %}
```

OleVik: Indeed, that works as well. Pushed to branch.

PR #1785, opened 2019-11-12
On admin save - clear media cache of the current saved page

buzatuAda:
On admin save - clear media cache of the current page in order to avoid waiting for 10 minutes for the cache to clear by itself.
I propose this because in the admin plugin when the media picker is opened there are 2 cases:
    - if the media files for a certain route aren't cached, they get cached for 10 minutes
    - if the media files for a certain route are already cached, it doesn't look anymore for new ones and just serves from cache the results that are already there.
Because of the second case if you upload new files into a page, you have to wait 10 minutes before they become available in the mediapicker

### DIFF
--- a/classes/adminbasecontroller.php
+++ b/classes/adminbasecontroller.php
@@ -1091,6 +1091,31 @@ class AdminBaseController
     }
 
     /**
+     * On admin save - clear media cache of the current page in order to avoid waiting for 10 minutes for the cache to clear
+     * by itself.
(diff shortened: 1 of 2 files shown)


mahagr:
FYI: If you try Flex Pages in upcoming 1.7, they already clear the media cache for the saved page. :)

buzatuAda:
> FYI: If you try Flex Pages in upcoming 1.7, they already clear the media cache for the saved page. :)

Oh ok I see. That's great! Then I'll use a workaround until then. Thanks

PR #1809, opened 2020-01-17
added on_demand option for filepicker

ricardo118:
on_demand: true option for filepicker field allows for users to only load the filepicker when the select field is clicked on. This is very useful for pages with many filepickers or large media files, as the page speed improves significantly


rhukster:
Please provide a PR for docs so we don't lose this new setting.  Cheers!

Issue #1824, opened 2020-02-03
Incorrect permissions display


lufog:
![image](https://user-images.githubusercontent.com/10379487/73695744-c7b4df80-46eb-11ea-8607-24a3c74d5f37.png)

rhukster:
Now that's interesting, seems we have some escaping enabled on our side that you are not seeing.  Will dig into this.

PR #1833, opened 2020-02-18
Fix incorrect permissions display

lufog:
Fix for #1824


mahagr:
That's odd, maybe `|trim` somehow makes twig to think that the macro output needs to be escaped in some cases. It works just fine for me without `|raw` and I always have autoescape turned on even in the site.

PR #1883, opened 2020-05-09
Feature/new scrollbars

rhukster:
Gemini scrollbars are pretty complicated and require a lot of tweaks and updates to work properly.  As part of the admin white-label effort, I've removed Gemini scrollbars, in favor of a lightweight more 'native' solution with [SimpleBar](https://grsmto.github.io/simplebar/) for the primary window + nav scrollbars as these work best when they are hidden by default and activate when needed. I've removed much of the JS that was handling updating Gemini, as these are no longer needed with SimpleBar.

All the other scrollbars use native styling for a simple thin 'bar' approach that is styled to match the current `content` area. This works much better for dark themes while still being 100% native requiring no JS.


themes/grav/app/forms/fields/colorpicker.js
```diff
@@ -82,24 +82,20 @@ export default class ColorpickerField {
         this.wrapper.addClass('cp-visible');
         this.updateFromInput();
 
-        let mainContainer = $('#admin-main .content-wrapper').data('scrollbar').getViewElement();
         this.wrapper.on(MOUSEDOWN, '.cp-grid, .cp-slider, .cp-opacity-slider', this.bound('bodyDown'));
         body.on(MOUSEMOVE, this.bound('bodyMove'));
         body.on(MOUSEDOWN, this.bound('bodyClick'));
         body.on(MOUSEUP, this.bound('targetReset'));
-        $(mainContainer).on('scroll', this.bound('reposition'));
```
w00fz, line 90: This is fine removing but you should add it back for the window scroll event. It is meant to auto-reposition the color picker popup when the page scrolls. Above removes completely that functionality.

themes/grav/app/forms/fields/colorpicker.js
```diff
@@ -82,24 +82,20 @@ export default class ColorpickerField {
...
     hide() {
         if (!this.built) { return; }
         this.wrapper.removeClass('cp-visible');
 
-        let mainContainer = $('#admin-main .content-wrapper').data('scrollbar').getViewElement();
         this.wrapper.undelegate(MOUSEDOWN, '.cp-grid, .cp-slider, .cp-opacity-slider', this.bound('bodyDown'));
         body.off(MOUSEMOVE, this.bound('bodyMove'));
         body.off(MOUSEDOWN, this.bound('bodyClick'));
         body.off(MOUSEUP, this.bound('targetReset'));
-        $(mainContainer).off('scroll', this.bound('reposition'));
```
w00fz, line 102: Same as above

themes/grav/app/main.js
```diff
@@ -61,10 +48,6 @@ export default {
     Dashboard,
     Pages,
     Forms,
-    Scrollbar: {
-        Scrollbar,
-        Instance: contentScrollbar
-    },
```
w00fz, line 67: I would leave this as empty objects for backward compatibility. Less prone to js failing.

PR #1905, opened 2020-06-16
onAdminAfterDelMedia provides affected media

drzraf:
When listening on the `onAdminAfterDelMedia` event, we are likely to be more interested by the media itself (eg, to prune caches), than the parent page (which may contain multiple medias).

Passing it won't hurt (and will likely help).

### DIFF
--- a/classes/admincontroller.php
+++ b/classes/admincontroller.php
@@ -1878,7 +1878,7 @@ class AdminController extends AdminBaseController
 
         $page = $this->admin->page(true);
         if ($page) {
-            $this->grav->fireEvent('onAdminAfterDelMedia', new Event(['page' => $page]));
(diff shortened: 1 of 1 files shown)


ricardo118:
wouldnt this send the path to a file that no longer exists?

drzraf:
Yes, but that way, the HTTP cache could be cleaned and, more importantly, if another file is uploaded, the cache of the old one won't be served
NB: I could not see `onAdminAfterAddMedia` being run

mahagr:
Needs a fix in Flex Objects, too.

mahagr:
Looks like there was another bug in the logic -- the page was always pointing to the home page!

I also took the liberty to split the new parameter into media object and filename.

PR #1922, opened 2020-07-08
Add an Event before a task is executed.

arkhi:
This type of event could be useful to prevent the execution of a task under certain conditions, processing the data…

I’m not entirely sure it is the best place for triggering the event but I hope you get the idea; I’d be happy to update the PR. :)


rhukster:
Moving this event could change current behavior and as such cannot be accepted.

Issue #1185, opened 2017-07-30
[Suggestion] Enable folder numeric prefix by default on module


paulmassen:
Hi!

Correct me if I'm wrong, but I thought in most cases, modular children are ordered manually with the handy ordering field on options tab (or advanced tab?).

Problem is we have to open every module to enable folder numeric prefix before being able to do so.
Don't you think it could be useful to have it enabled by default?

Thanks again,

Paul

RomainFallet:
What about this ? I'm having the same issue and setting a default value for ordering in blueprint does not work.

Kangouroops:
Helo, 

Same here, adding a default value won't work.

Amaury

carlu93:
This would be really handy. :)

karolzyk:
Plus+ on this!

RastaTux:
We have this problem as well and it's very annoying to tell an editor to enable this field everytime...

RastaTux:
Actually could we change that from [Suggestion] to bug, as of the default value of the blueprint gets ignored here?

dithom:
Same here, custom blueprint doesn't override the default value, should be a bug

stephan-strate:
Same here, default value does not work!

RastaTux:
Hey @rhukster,
May you have a look at this issue? It's titled a [Suggestion], but this seems more than a bug that happened to many users. :)

Sogl:
+1 to add this functionality

hotdoy:
That's a step I could skip too.
I vote yes!

junglegobs:
+1 to add this!

Schw3pps:
+1 to add this functionality

PR #1938, opened 2020-07-28
Allow enabling folder numeric prefix by default

TheoAcker12:
Related to issue #1185: This allows enabling folder numeric prefix by default for pages with no siblings. While the related issue is only about modular pages, this option applies to all types, as this is a constant annoyance for me. It seems like it can't hurt to at least have an option in the plugin.

By default, the new option is set to false, since that is currently how the plugin operates.

I am happy to update the PR if changes need to be made before merging it.

### DIFF
--- a/admin.yaml
+++ b/admin.yaml
@@ -31,6 +31,7 @@ edit_mode: normal
 frontend_preview_target: inline
 show_github_msg: true
 pages_list_display_field: title
+enable_numeric_prefix: false
 google_fonts: false
 admin_icons: line-awesome
 enable_auto_updates_check: true
--- a/blueprints.yaml
+++ b/blueprints.yaml
@@ -211,6 +211,18 @@ form:
       label: Pages List Display Field
       help: "Field of the page to use in the list of pages if present. Defaults/Fallback to title."
 
+    enable_numeric_prefix:
+      type: toggle
+      label: Enable Folder Numeric Prefix
+      highlight: 1
+      default: 0
+      options:
+        1: PLUGIN_ADMIN.ENABLED
+        0: PLUGIN_ADMIN.DISABLED
+      validate:
+        type: bool
+      help: "Enables folder numeric prefix for sortable pages by default if a new page has no siblings."
+
     enable_auto_updates_check:
       type: toggle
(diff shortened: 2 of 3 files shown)


mahagr:
I don't agree with this change as you can already create new page template with ordering turned on or override the existing one. 

FYI: This one would need an implementation also in Flex Pages side.

Maybach91:
@mahagr  how can i override it? i dont find the correct key for it since `ordering: true` is not working within the advanced field 
```

form:
  fields:

    tabs:
      fields:
        advanced:
          fields:

            ordering: true
```

mahagr:
Override `user/plugins/admin/blueprints/admin/pages/modular_new.yaml` with your own version to change the default behaviour when creating a new modular page. You can put the copied file to your main blueprint folder: `user/blueprints/admin/pages/modular_new.yaml`. Add:

```yaml
    visible:
      type: hidden
      default: 1
      validate:
        type: bool
```

into the file and modular pages should always use the numeric prefix.

drzraf:
Elegant solution, but even after reread https://learn.getgrav.org/16/forms/blueprints/advanced-features I think an example of the **minimum** possible such override would be good to post:

Out of my mind (untested and probably wrong):
```yaml
'@extends': default

form:
  fields:
    import@1:
        type: admin/pages/modular_new
        context: 'blueprints://'
    visible:
      type: hidden
      default: 1
      validate:
        type: bool
```

mahagr:
I asked to copy the file and add the field. :)

sque89:
@mahagr one inconvinience with this solution is the lack of update safety. If this yaml file will be changed in a future version of the admin plugin i need to copy the file again and change it again according to my needs correct?

rcheetah:
What would be the way to realize this in 1.7 with flex pages? The solution above does not work for me, but @mahagr mentioned „FYI: This one would need an implementation also in Flex Pages side.“

mahagr:
Using blueprint should work in Flex, just not this commit.

TheoAcker12:
It sounds like this is obsolete with Flex. I'm going to go ahead and close this.

PR #2008, opened 2020-12-29
Make upstream promotions' notifications optional

drzraf:
And incidentally, avoid servers being tracked by that endpoint (https://getgrav.org/notifications.json)

### DIFF
--- a/classes/admin.php
+++ b/classes/admin.php
@@ -1337,7 +1337,7 @@ class Admin
         $notifications = $notifications_content['data'] ?? array();
         $timeout = $this->grav['config']->get('system.session.timeout', 1800);
 
-        if ($force || !$last_checked || empty($notifications) || (time() - $last_checked > $timeout)) {
+        if ($this->grav['config']->get('plugins.admin.upstream_notifications', true) && ($force || !$last_checked || empty($notifications) || (time() - $last_checked > $timeout))) {
             $body = Response::get('https://getgrav.org/notifications.json?' . time());
 //            $body = Response::get('http://localhost/notifications.json?' . time());
             $notifications = json_decode($body, true);


w00fz:
hey @drzraf I see what you are trying to do and this will work, however you are missing some elements.

For starter, this should be a 1.10 change, right now you are proposing it to develop and I am not sure this is worth for us spend time in merging, there have been some changes in the partials and how notifications work and with Grav 1.7 and Admin 1.10 close to stable release, I don't feel like adding new features that potentially can break the current 1.6/1.9 stable release.

Furthermore, this should also be a toggle in admin, the blueprints already have the notification settings, you should add your new `upstream_notifications` at the top of the notifications section.

I would also call this `notifications_enabled`, once you disable the fetching of new notifications, you are de-facto disabling the admin notification feature, which makes this naming more appropriate.

This brings to the last of the issues which is making sure your new settings apply to the rest of the admin. So far our notifications locations are independent, your setting will make this sort of unified. This implies that what you have alone will potentially break other part of the admin.

1. [javascript-config.html.twig](https://github.com/getgrav/grav-plugin-admin/blob/1.10/themes/grav/templates/partials/javascript-config.html.twig#L2): A check needs to be added here. (ie, `config.plugins.admin.notifications.notifications_enabled and (...rest..)`
2. Because there are different [partials](https://github.com/getgrav/grav-plugin-admin/tree/1.10/themes/grav/templates/partials) for the notifications, please make sure your changes also don't affect the rest of admin, particularly dashboard/plugins/themes which is where it's possible to have these. The feed also is affected by notifications in dashboard.

If you can make all these changes we'll gladly merge this in (feel free to also update the CHANGELOG).

Cheers!

drzraf:
As you probably understood, I found that latest promotional banner (forcefully appearing everyone's backend) somehow... intrusive for editors.
I added that patch to the installations I manage but I'm sorry I can't afford the time to make it more refined to be production grade as you suggested.

As a suggestion, you may want to reserve this kind of banner to a couple of specific users (admin role) instead of the (possible dozens of) editors using a given Grav instance.

w00fz:
You can disable the notification by turning off plugins/themes/dashboard notifications from the admin settings if that’s your intention, no need to hack the code base. It’s always been there as an option.

I’m sorry if you felt this intrusive, we made it so that notifications can be hidden once viewed and we have options to let you completely disable them from appearing in the admin. We never really pushed any banner like this all these years even if we could have promoted our trilby services, but for Grav Premium being completely related to Grav, the ecosystem and the community, wether admin or editors, we felt appropriate utilizing our notifications to reach out to everyone.

As this PR stands unfortunately I cannot merge it because it requires some work as stated above and not really what we are looking for since all you really want is to disable the notifications which you can already do. Your idea of targeting certain type of users only is a good one we might consider in the future once our focus shifts from the imminent release of Grav 1.7 and Admin 1.10.

Thanks for the feedbacks, they are appreciated.

OleVik:
Aside from vehemently opposing the addition of call-home scripts that both track the user and create an additional channel for side-loading malicious scripts, the current notification-settings are insufficient @w00fz. Feeds, Dashboard, Plugins, and Themes do not cover advertising.

w00fz:
> Aside from vehemently opposing the addition of call-home scripts that both track the user and create an additional channel for side-loading malicious scripts, the current notification-settings are insufficient @w00fz. Feeds, Dashboard, Plugins, and Themes do not cover advertising.

It hasn't been a problem all these years when you were getting feed notifications blog posts and releases updates. Those have been coming to you the same way we are delivering this other notification about Grav Premium. Nothing malicious about what we are doing, just trying to get the word out.

You have all the rights of wanting it disabled and there are settings to do so. You must be doing something wrong because this works for me:

CleanShot 2021-01-20 at 12 43 43@2x

OleVik:
I already block those. But the Advertisement is not equal to Feeds, Dashboard, Plugins, and Themes, it should be an additional setting.

w00fz:
Sure if we'll ever do advertisements that's a good idea. However, this is just a notification.

OleVik:
It's a link to a paid product, an advertisement by any name. Third-party direct-access to in-product advertisements would be even worse.

PR #2024, opened 2021-01-20
Allow blueprint creation modals in other pages

stephan-strate:
I have added the possibility to set a form action for the blueprint partials.
This way you can eg. add a "Add page" button to another admin page and define the form being sent to `admin_route('pages')`. If no `form_action` parameter is passed, the value is empty and the form will be sent to self.

### DIFF
--- a/themes/grav/templates/partials/blueprints-new-folder.html.twig
+++ b/themes/grav/templates/partials/blueprints-new-folder.html.twig
@@ -1,7 +1,7 @@
 {% set form_id = form_id ? form_id : 'blueprints' %}
 {% set scope = scope ?: 'data.' %}
 
-<form action="{{ form_action }}" id="{{ form_id }}" method="post" data-grav-form="{{ form_id }}" data-grav-keepalive="true">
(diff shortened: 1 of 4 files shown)


themes/grav/templates/partials/blueprints-new-folder.html.twig
```diff
@@ -1,7 +1,7 @@
 {% set form_id = form_id ? form_id : 'blueprints' %}
 {% set scope = scope ?: 'data.' %}
 
-<form id="{{ form_id }}" method="post" data-grav-form="{{ form_id }}" data-grav-keepalive="true">
+<form action="{{ form_action }}" id="{{ form_id }}" method="post" data-grav-form="{{ form_id }}" data-grav-keepalive="true">
```
mahagr, line 4: HTML 5 specs do not allow empty action. To avoid any possible problems (though browsers do work with it), I would remove the attribute when it's not being used.

mahagr (changes requested):
Can you fix this:

themes/grav/templates/partials/blueprints-new-folder.html.twig
stephan-strate: @mahagr Sure, done!

PR #2055, opened 2021-02-03
Add item at the top if top button is clicked in list field

Karmalakas:
Pretty much title says it all. IMO it makes much more sense, that top button would add item at the top, because if list is long enough, you are not even seeing if something happens when item is added at the bottom. Created this PR because of [the question asked in forum](https://discourse.getgrav.org/t/blueprint-type-list-new-data-on-top-of-array/15254). Also created same change [PR for Flex Objects plugin](https://github.com/trilbymedia/grav-plugin-flex-objects/pull/105)

So the solution would be either like this PR suggests, or button actions should be configurable via Admin config somewhere.

### DIFF
--- a/themes/grav/templates/forms/fields/list/list.html.twig
+++ b/themes/grav/templates/forms/fields/list/list.html.twig
@@ -56,7 +56,7 @@
                         <button class="button{{ not value|length ? ' hidden' : '' }}" type="button" data-action="sort" data-action-sort="{{ field.sortby }}" data-action-sort-dir="{{ field.sortby_dir|default('asc') }}"
                                 {% if field.disabled or isDisabledToggleable %}disabled="disabled"{% endif %}><i class="fa fa-sort-amount-{{ field.sortby_dir|default('asc') }}"></i> {{ btnSortLabel|e|tu }} '{{ field.sortby }}'</button>
                     {% endif %}
-                    <button class="button" type="button" data-action="add" data-action-add="bottom"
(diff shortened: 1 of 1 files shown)


w00fz:
This is a bit of an impactful change to just merge in as is. The expected behavior would change and although I totally see your point I also think it’s a matter of preference and quite definitely contextual. Sometimes I might want it added at the top, sometimes at the bottom.

I think a better solution is to introduce a `placement` option so that via blueprint, one can decide if for a particular list the new entry should be added at the top or at the bottom.

Something like `data-action-add="{{ field.placement|default('bottom') }}"` with bottom as default to ensure backward compatibility but now customizable from the blueprint to allow too.

Can you make this change on both PRs? Also I didn’t test this so make sure you do.

Karmalakas:
Added field `placement` property check.

If `placement` value is `top` or `bottom`, both buttons add item to top or bottom respectively.
If `placement` value is `position`, item is added depending on a clicked button position - if top button is clicked, item added to top and if bottom button, item added to bottom.
  
Defaults to `bottom`

Karmalakas:
Also updated [Flex Objects plugin PR](https://github.com/trilbymedia/grav-plugin-flex-objects/pull/105)

Karmalakas:
Created [PR for docs](https://github.com/getgrav/grav-learn/pull/901)

w00fz (approved):
I like this thanks, I approve.

PR #2117, opened 2021-04-13
Fix side-tabs navigation in page edit

PixelT:
By default, tabs-nav as side-tabs on edit page has redundant margin (this margin is added to all tabs-nav, to have space for `Normal | Expert` buttons on the right side in top - horizontal tabs-nav), but for side-tabs it's not needed.
This PR fix this issue.

image


mahagr:
@rhukster Can you take a look on this?

rhukster:
looks ok to me.. i'll merge and fix if anything comes up.

Issue #2127, opened 2021-04-20
Image Button in Editor Shouldn't Start the http://


McLaynV:
I suggest the same change as https://github.com/getgrav/grav-plugin-admin/issues/813 for `image` 
https://github.com/getgrav/grav-plugin-admin/blob/d02f3a12e204708843024ce2ca9623fb6e152646/themes/grav/app/forms/fields/editor/buttons.js#L245 
Another good reason is that HTTPS is getting more common than HTTP.

mahagr:
@w00fz There's no such thing as `image:` URI, which is why there's no need to handle it as a special case.

w00fz:
@mahagr not sure what you mean by `image:`, this is not a special case.

I think @McLaynV suggestion is to just remove the automatic rendered prefix of `http://` when adding an image. I fully agree with this since it was poorly assuming your image URL was going to be remote AND not using SSL.

This is fixed now.

McLaynV:
Hello @w00fz, there is one more occurrence in 
https://raw.githubusercontent.com/getgrav/grav-plugin-admin/develop/themes/grav/js/app.js

w00fz:
@McLaynV i believe that file is a leftover from before my cleanup in Nov 2020. I removed it.

Issue #813, opened 2016-10-06
Link Button in Editor Shouldn't Start the http://


FrugalGeek:
I noticed that if I click the link button it generates: `[Linked Text](http://)`. If I'm copying and pasting a URL from another source this creates an extra step as I have to go back and delete the placed `http://`.

Very rarely does someone writing an article or a blog post want to write the url from scratch. More often than not, they'll have a direct URL already copied and ready to be pasted.

So, I suggest having the link button instead create `[Linked Text]()` and place the cursor between the `()`.

flaviocopes:
Same works for the image button. I tend to agree. @w00fz what you think?

w00fz:
I'm ok changing this I just hope it won't be confusing because it does look a bit weird when it's only `[Linked Text]()` :)

PR #2128, opened 2021-04-20
Don't prefix image URLs with http in editor

McLaynV:
Does https://github.com/getgrav/grav-plugin-admin/issues/2127

It is the same change as https://github.com/getgrav/grav-plugin-admin/issues/813 - just for `image` button
Another good reason is that HTTPS is getting more common than HTTP.


mahagr:
@w00fz For you.

McLaynV:
Fixed in https://github.com/getgrav/grav-plugin-admin/commit/084002bd77311edf224e5ed2a93c7c8f81e02479

w00fz:
Sorry didn't realize you had opened a PR.

Issue #2202, opened 2021-12-08
Remove security and privacy issue Gravatar


Rotzbua:
## Current reason

Gravatar is not a reliable service. There was a data leak where more than 100 million(!!) emails were leaked.

References:
* https://monitor.firefox.com/breach-details/Gravatar
* https://haveibeenpwned.com/

Gravatar is a security risk and privacy issue. It should be removed or disabled by default in my opinion.

## Other reasons

* It also causes problems when users want to use new security features such as CSP.
* In offline mode it causes display errors. #1335
* May violate data protection laws or policies, which must additionally be evaluated by the user.
* Requested years ago #650

w00fz:
I left a comment in the PR with a bunch of changes request to make your PR valid. However, like I mentioned, I'm failing to see how this can be considered a security and/or privacy concern, considering the only way you can see this Gravatar is if you are logged in.

No one outside of the Admin can access the Gravatar and anyone within the admin, that can see the Gravatar, can already see the email address, so there really isn't a privacy concern considering the email is already there out in the wild.

That said, I like the idea of this being an opt-in solution and I gave you some directions in the PR because what you have right now can't really work.

Cheers

atomGit:
this is crazy - this issue [goes back at least 6 years](https://github.com/getgrav/grav-plugin-admin/issues/339) and the default is _still_ to use a "free" data-harvesting, privacy-disrespecting 3rd party service to generate a useless image???

this says a lot about the lack of ethics of the developer(s) responsible for this nonsense

PR #2211, opened 2021-12-27
Remove security and privacy issue Gravatar #2202

Rotzbua:
Fixes: Remove security and privacy issue Gravatar #2202

Change: Replaces Gravatar image with an user icon.


w00fz:
I'm failing to see how this can be considered a privacy concern, in this context specifically. 
We are talking about a use of Gravatar that is purely visible to the user itself, within the admin, when logged in. This is not shared or visible anywhere else, outside your logged in user. If you can see this user Gravatar, you can already see the user email and change password.

I would be ok making this an opt-in solution, but I did checkout your PR and it's not quite ready for a merge. Here the issues I see:
CleanShot 2022-02-04 at 10 07 51@2x

1. This seems to be still using Gravatar
2. This should be consistent with the size of the avatar view, 200px font-size perhaps
3. If you want this to make it opt-in and disabled by default, we do need an option to change this behavior, right now I am not entirely sure what the PR is supposed to do but it looks like it's missing something.

Please make those changes and we can integrate this

froschdesign:
@w00fz 
> I'm failing to see how this can be considered a privacy concern, in this context specifically.

The Gravatar service itself is the problem. The reference was given in the related issue report: https://monitor.firefox.com/breach-details/Gravatar

> 3. If you want this to make it opt-in and disabled by default…

Regardless of security issues, the use of external services should **always** be optional.

themes/grav/scss/template/_admin.scss
```diff
@@ -1114,7 +1114,7 @@ body.sidebar-quickopen #admin-main {
         linep-height: 1.2;
     }
 
-    .gravatar {
+    avatar-upload-hint {
```
Rotzbua, line 1117: dot missing

themes/grav/scss/preset.scss
```diff
@@ -1286,7 +1286,7 @@ form {
         }
     }
 
-    .gravatar {
+    avatar-upload-hint {
```
Rotzbua, line 1289: dot missing

Rotzbua:
>     1. This seems to be still using Gravatar

Left avatar template is:
```twig
{% set user_avatar = admin.user.getAvatarUrl() %}
{% if user_avatar %}
```
Right avatar template is:
```twig
{% if data.avatar %}
    <label><img src="{{ data.getAvatarUrl() }}" /></label>
```

Indeed this is a bug in the plugin. The left template does it correctly. The right template is wrong, because it only checks for local avatar by `data.avatar`. But grav also allowes external avatar which should be requested by `getAvatarUrl` as documented in [core](https://github.com/getgrav/grav/blob/163cd5a03f0d9713a3b8a529dd418d44352f2614/system/src/Grav/Common/User/Traits/UserTrait.php#L123) :

>     * Note: if there's no local avatar image for the user, you should call getAvatarUrl() to get the external avatar URL.


So you are right that both templates should behave similar. I changed them in the PR so they just relay on `getAvatarUrl`.

>     2. This should be consistent with the size of the avatar view, 200px font-size perhaps

I removed the fallback. A non existing avatar does not destroy the layout. If there should be a default image it should be served by core functions.

>     3. If you want this to make it opt-in and disabled by default, we do need an option to change this behavior, right now I am not entirely sure what the PR is supposed to do but it looks like it's missing something.

I reworked the PR. Now it does not change the behavour, because `getAvatarUrl` serves by default a fallback avatar from gravatar. So currently there should be no beak until `getAvatarUrl` is changed.

mahagr:
I prefer this one:

https://github.com/getgrav/grav/pull/3550

rhukster:
See: https://github.com/getgrav/grav/pull/3551

PR #2235, opened 2022-02-10
Ricardo118 patch collection

ricardo118:
fix a bug for the new elements field not showing/hiding properly within lists


w00fz:
Cheers

PR #2313, opened 2022-09-05
Update fieldset.html.twig

iainsgillis:
Fix `fieldset.text` not rendering when `markdown` isn't set, or is falsy.


rhukster:
Looks like a legit bug, thanks.

iainsgillis:
Glad to help -- thanks for all of your work.

PR #2376, opened 2023-10-07
Fixed some 'medium' tags in select fields and fixed fa-question-circle icon in blueprint with markdown option and some text in help option

pmoreno-rodriguez:


rhukster:
thanks

PR #2424, opened 2024-10-20
Fixed typo in classes and outerclasses and added outerclasses to column and section fields.  New div block for field.outerclasses. New headgin_classes tag for h1 tag

pmoreno-rodriguez:


pmoreno-rodriguez:
Hi @rhukster

Regarding this pull request, I have modified the following to make it consistent across all Grav theme templates in the Admin plugin:

Fixed field_classes to field.classes for some field types.
Separated classes from outerclasses, adding a div block for the latter.
Added field.heading_classes to differentiate classes in the h1 tag from classes used in other blocks
I am still reviewing the use of classes in the rest of the Admin plugin templates. If you see these changes well, I will add the necessary help in Grav Learn Docs for the new attributes

pmoreno-rodriguez:
Ok. @rk

I think I've made too many commits on the same files, in two different pull requests, and maybe I should organize my Admin plugin repository, and make a single pull request with all the files I propose to modify. That way it would be clearer.

Right now, even I'm getting confused. 

Do you think I should close pull requests #2432 and #2424 to unify all the changes in a single pr?

themes/grav/templates/forms/field.html.twig
```diff
@@ -25,9 +25,14 @@
...
 {% set isDisabledToggleable = toggleable and not toggleableChecked %}
+{# Backwards compatibility with field.outer_classes instead of field.outerclasses #}
+{% set outerclasses = field.outerclasses ?? field.outer_classes %}
 
 {% block field %}
-    <div class="form-field grid{% if vertical %} vertical{% endif %}{% if toggleable %} form-field-toggleable{% endif %} {{ field.outerclasses }} {{ field.field_classes }}">
+    {% if outerclasses %}
+        <div class="{{ outerclasses }}">
+    {% endif %}
+    <div class="form-field grid{% if vertical %} vertical{% endif %}{% if toggleable %} form-field-toggleable{% endif %} {{ field.classes }}">
```
Karmalakas, line 35: This completely changes HTML of the field (same goes for section below) 🧐
The way I see it, it's a breaking change

themes/grav/templates/forms/fields/column/column.html.twig
```diff
@@ -1,8 +0,0 @@
-{% extends "forms/field.html.twig" %}
```
Karmalakas, line 1: Where did the column go? I don't see any replacement 👀

Karmalakas (changes requested):
It's up to Grav team, but I would not approve this PR 😔

pmoreno-rodriguez:
> It's up to Grav team, but I would not approve this PR 😔

What do you recommend me?

Karmalakas:
> > It's up to Grav team, but I would not approve this PR 😔
> 
> What do you recommend me?

I can't really recommend much, because PR description is empty, and I don't understand why this change is actually needed in the first place 😕 I see in the comment you listed what you did, but what's the reason for this? 🤔

themes/grav/templates/forms/fields/column/column.html.twig
pmoreno-rodriguez: You're right. That was a big mistake.

themes/grav/templates/forms/field.html.twig
pmoreno-rodriguez: There is no point in this line:
```
<div class="form-section {{ field.field_classes }} {{ field.outer_classes }}">
```
having field_classes and outer_classes applied to the same div block. Either apply outer_classes to a different block, which wraps the next one (my choice), or remove outer_classes from that div tag.

pmoreno-rodriguez:
> > What do you recommend me?
> 
> I can't really recommend much, because PR description is empty, and I don't understand why this change is actually needed in the first place 😕 I see in the comment you listed what you did, but what's the reason for this? 🤔

The PR is not well structured. I'm going to do it differently, with small PRs, each with its proposed changes.
The reason for these changes is to make the Grav documentation consistent with the code. In the [forms documentation](https://learn.getgrav.org/17/forms/forms/fields-available), you can see that `classes` and `outerclasses` are proposed as common attributes. In some Admin and Form plugin files, some of these attributes are missing or not well defined.

themes/grav/templates/forms/field.html.twig
Karmalakas: Imagine I have in my CSS:
```css
.form-field.grid.outer-class { color: red; }
```

Old HTML, which shows text in red:
```html
<div class="form-field grid {{ field.outerclasses }}"> ... </div>
```

New HTML, which doesn't show text in red any more:
```html
<div class="{{ outerclasses }}">
    <div class="form-field grid"> ... </div>
</div>
```

pmoreno-rodriguez: I understand from what you put that in the blueprint you have put:
`outerclassses: outer-class`

Well, now imagine that you want to set the red color for each field inside the outerclasses block, that is, something like this:
```
<div class="{{ outerclasses }}">
<div class="form-field grid {{ field.classes }}"> ... </div>
</div>
```
Now your red text is applied to field.classes, and in outerclasses, for example, you could set a padding or a margin between the fields.

With the current code, this cannot be done.
Let me finish some changes and I will show you what I am telling you. I am modifying the Deliver theme and in it I am implementing some of these changes.

pmoreno-rodriguez:
I'm closing this PR to recreate it with corrections.

PR #2432, opened 2024-12-02
Changed `field.classes` on some field types.

pmoreno-rodriguez:


pmoreno-rodriguez:
Hi @rhukster 

Regarding this pull request, I have modified the following to make it consistent across all Grav theme templates in the Admin plugin:
1) Fixed field_classes to field.classes for some field types.
2) Separated classes from outerclasses, adding a div block for the latter.
3) Added field.heading_classes to differentiate classes in the h1 tag from classes used in other blocks

I am still reviewing the use of classes in the rest of the Admin plugin templates. If you see these changes well, I will add the necessary help in Grav Learn Docs for the new attributes

rhukster:
Just fyi the biggest issue is consistency between the structure of this fields in the form plugin and the admin plugin. Admin uses both fields from its own location as well as those from the form’s, while front end only has access to form’s fields.

pmoreno-rodriguez:
So, I understand that I should find a way to standardize both the fields of the form plugin and the admin plugin, so that both use the same class attributes in the same types of fields, while including those common attributes in the Grav documentation. Am I right? If so, I will spend time searching in all the templates of both the Form and Admin plugin to try to standardize the common attributes in both

pmoreno-rodriguez:
I'm closing this PR to recreate it with corrections.

Issue #1479, opened 2018-07-04
Grav Admin login page doesn't obey force_ssl


brendanpetty:
After enabling the force_ssl option in Grav system config, I've found it doesn't redirect the /admin page to a https connection. This is the most important page to have forced to redirect.
Either the admin plugin needs to obey Grav's system setting for this, or it needs its own setting to force https on the login (and refuse to accept a password if not https)

brendanpetty:
Thanks!

PR #2435, opened 2025-01-03
fix(force_ssl): use scheme instead of server https index

pips-:
using some proxy / load balancer (openbsd relayd -> httpd -> php-fpm) may not properly set server https index, thus leading to infinite redirections on force_ssl parameter set to true

align https detection with grav-core code:

https://github.com/getgrav/grav/blob/a1c116dd82c761ecf9800e412fcd9159722ad769/system/src/Grav/Common/Service/PagesServiceProvider.php#L72-L78

slightly related to #1479 (came across this portion of code by searching using force_ssl keyword and stumbled upon this issue)

### DIFF
--- a/admin.php
+++ b/admin.php
@@ -396,9 +396,10 @@ class AdminPlugin extends Plugin
 
         // Force SSL with redirect if required
         if ($config->get('system.force_ssl')) {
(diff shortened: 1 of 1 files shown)


rhukster:
Thanks.

PR #2438, opened 2025-01-07
Added classes for fields, which are missing in some field types

pmoreno-rodriguez:


pmoreno-rodriguez:
`classes` is a common attribute for fields, according to [Grav Documentation](https://learn.getgrav.org/17/forms/blueprints/fields-available#common-fields-attributes).
The fields of this PR have not the classes attribute.

pmoreno-rodriguez:
Hi @rhukster . Any advance about this PR?

PR #2445, opened 2025-03-03
Support in Admin plugin to upload Fontawesome 6 icons using Iconpicker

pmoreno-rodriguez:
Hi.
I've made some changes to the admin plugin to support the latest Fontawesome 6 version for Iconpicker buttons, which until now were displaying icons from Forkawesome, which was based on Fontawesome 4.7, and quite outdated.

Do you think it's feasible to add this functionality to Grav for its upcoming 1.8 release?

I'll give you a couple of screenshots

fontawesome_v6b
fontawesome_v6a


rhukster:
Is this the free stuff only? If we were to do this, we should probably upadte the icons in the Admin plugin also.  The only issue is that there was some additional items in forkawesome that were added after 4.7.

pmoreno-rodriguez:
In fact this PR is for the Admin plugin, and includes Fontawesome fonts version 6.

You can try it yourself if you want in a clean Grav installation. 

Obviously, the creators of themes and plugins should adapt their code, because in Fontawesome 6 there is more than one prefix, not just `fa`. I'm studying this

PR #2485, opened 2025-12-06
Fix admin recovery crash when blueprints.fields is null

onlsol:
### Summary

Admin can go into recovery mode with:

> foreach() argument must be of type array|object, null given

when rendering blueprints where `blueprints.fields` is `null`.

### Steps to reproduce

1. Use a blueprint that does not define any `form/fields`, so `blueprints.fields` ends up `null`.
2. Open the corresponding page in the Admin panel.
3. Admin switches to recovery mode with an exception pointing at [partials/blueprints.html.twig](cci:7://file:///home/rabby/Desktop/prj/ay/artf.ai/user/plugins/admin/themes/grav/templates/partials/blueprints.html.twig:0:0-0:0) line 7.

### Technical details

In the Admin plugin there are two twig templates that call `Admin::findFormFields()` like this:

```twig
{% if admin.findFormFields('file', blueprints.fields) %}


rhukster:
Bad pull request.. 47 files changes can't be correct

Issue #2468, opened 2025-10-07
Grav System Log Output date


teixeiradacunha:
The application event log (Grav System Log Output) is showing an issue: all entries display the same date, which always corresponds to the most recent event.

<img width="585" height="660" alt="Image" src="https://github.com/user-attachments/assets/c10571cb-aad1-4e5a-9b93-6cdf62cee6a7" />

Xoriander:
The problem was reproducible for me with Grav 1.8.0-beta.28.
Grav uses the Monolog logger and the datetime format was changed https://github.com/Seldaek/monolog/commit/45b5e0e33e5b95f82d7b99b304ae20ff26e31f3f this could be the reason.

To resolve it I changed:

https://github.com/getgrav/grav/blob/7a6b8a90d454dbc42396e6be84700a61297076a0/system/src/Grav/Common/Helpers/LogViewer.php#L142-L144

to:

```
    try {
        $date = new DateTime($data['date']);
    } catch (DateMalformedStringException $e)  {
        $date = null;
    }
    
    return [
        'date' => $date,

```
and:

https://github.com/getgrav/grav-plugin-admin/blob/1ec8fabc791216081a8987d2850888a2e3ca2af2/themes/grav/templates/partials/tools-logs.html.twig#L56

to:

`<td class="date" title="{{ log.date|date('r') }}">{{ log.date|date('D, d M Y H:i:s') }}</td>`

rhukster:
Thanks for tracking this down.  Why not create a PR and you'll get credit for the fix and make my life easier :)

PR #2487, opened 2025-12-10
Part of the fix for #2468 - Wrong log output date

Xoriander:
This pull request is part of the fix for getgrav/grav-plugin-admin#2468
The main part is this pull request getgrav/grav#4007

To show the date like the previous Grav versions the format 'D, d M Y H:i:s' was added.
To avoid to output the current date if none date was parsed from the logfile, an if statement was added.


rhukster:
yikes, clearly a mistake. closing

Xoriander:
I have created a new pull request #2497.
It's better to leave the date format from the twig filter empty as it was, because the date format comes from here:
https://github.com/getgrav/grav/blob/c191b0b47efb25fe28effe00dcebc623fffa0177/system/src/Grav/Common/Twig/Twig.php#L249

PR #2508, opened 2026-03-17
Added classes for fields, which are missing in some field types

pmoreno-rodriguez:


pmoreno-rodriguez:
Hi! Just following up on this PR in case it got lost in the queue.

This change adds the missing CSS classes to some field types, making their markup more consistent with the other fields and allowing them to be targeted reliably with CSS.

The PR is still ready for review. Whenever you have a chance, I’d appreciate a look at it.

Thanks!

rhukster:
Nice catch. `field.classes` is honoured by 16 other field templates and these three were the odd ones out, so this is a real consistency gap. Merging.

One bonus while we're on that line: `array.html.twig` was emitting a second `class="one-child"` attribute alongside the first one, so the browser silently dropped it and the server-rendered `one-child` never actually applied (the JS re-applies it at runtime, which is why nobody noticed). I've folded that into the same class string on top of your change.

PR #2520, opened 2026-09-13
fix: show top expand and collapse controls for list fields

### LINKED ISSUE #2503: [Bug] List field "Expand All / Collapse All" buttons not rendered in "top" position when using controls: both
In `forms/fields/list/list.html.twig`, the `collapsible` variable is computed inside the `{% for key, val in value %}` loop (line ~90), which means it is `undefined` at the time the `top` block of `collection-actions` is rendered.

**Result :** When using `controls: both` in a blueprint list field, the **"Expand All" and "Collapse All" buttons appear correctly at the bottom** of the list, but are **missing at the top**.

**To reproduce :**
```
yaml
  type: list
  style: vertical
  controls: both
  collapsed: true
  btnLabel: 'Add item'
  fields:
    .field_a:
      type: text
      label: 'Field A'
    .field_b:
      type: text
      label: 'Field B'

```
Expected : "Expand All", "Collapse All" and "+ Add item" buttons visible **both at top and bottom**.
Actual : Only "+ Add item" appears at top. "Expand All" / "Collapse All" are missing at top.

**Root cause :**
```
twig
{# In list.html.twig — collapsible is set here, INSIDE the for loop #}
{% for key, val in value %}
    {% set collapsible = field.fields|length > 1 and ... %}  {# ← too late #}
But the top block is rendered before this loop :
twig{% if fieldControls in ['top', 'both'] %}
    <div class="collection-actions">
        {% if collapsible %}   {# ← undefined here, evaluates to false #}

```

Fix :
Move the `collapsible` declaration before `{% block contents %}` :
```
twig
{% set fieldControls = field.controls|default('bottom') %}

{# FIX: compute collapsible before block contents so it's available in both top AND bottom collection-actions #}
{% set collapsible = field.fields is defined and field.fields|length > 1 and (field.collapsible is not defined or field.collapsible) %}

{% block contents %}
```

This one-line move makes `collapsible` available to both the `top` and `bottom collection-actions` blocks without changing any other behavior.

**Affected file :**
`user/plugins/form/templates/forms/fields/list/list.html.twig`
(also reproduced in `user/plugins/admin/themes/grav/templates/forms/fields/list/list.html.twig`)
**Grav version :** tested on 1.7.x
**Plugin form version :** tested on current

sridharkalaibala:
With `controls: top` or `controls: both`, the list field renders its top action group before `collapsible` is assigned, so Expand All and Collapse All are missing there. Compute this field-level value before the contents block so the top controls, list items, and bottom controls share it. Guard absent child fields, as the old assignment was inside the `field.fields` condition.

Fixes #2503 for the Admin template on develop.

Validation:
- Rendered the actual modified Twig template with PHP 8.3.6 / Twig 3.28.0 using a standalone harness. The parent field wrapper and child-field helpers were simplified; list contents/action markup came from the repository template.
- 54 combinations: top/bottom/both controls, zero/one/two child fields, default/false/true `collapsible`, and empty/populated values. Before: 16 missing-button assertions across 8 configurations. After: all cases pass; Add Item button counts remain correct.
- Existing JavaScript delegates both actions to direct `.collection-actions` groups in the same list wrapper, covering either placement; no JavaScript changed.
- `git diff --check` passes. No complete Grav installation, older Twig runtime, or browser interaction test was run.

The Form plugin's current develop tree has no `templates/forms/fields/list/list.html.twig`, so this patch targets the existing Admin template named in the report.

Prepared with AI assistance; the template diff and before/after render results were reviewed.


rhukster:
Thanks — hoisting the `collapsible` calculation to the top matches how `value`/`name`/`fieldControls` are already set in that template, and it fixes a latent nested-list case as a bonus. Merged.

PR #2521, opened 2026-09-13
fix: preserve zero-valued blueprint placeholders

sridharkalaibala:
An Admin blueprint text field with `placeholder: 0` or `placeholder: '0'` loses its placeholder because the base field template tests truthiness. Use an emptiness test so zero renders while missing, null, false and empty-string values remain omitted. Existing translation and autoescaping stay intact.

Addresses getgrav/grav-plugin-form#511 in the Admin renderer. The Form plugin has a separate frontend template and companion fix.

Validated the actual Admin text template and inheritance chain with Twig 3.28.0/PHP 8.3.6 and minimal Grav helper filters: missing/null/false/empty, integer/string/float zero, a nonzero number, text and quotes/ampersands/angle brackets. Original: 3 fail / 7 pass. Patched: all 10 pass. The same matrix passes for the patched frontend renderer (20 combined checks). Diff check passes. This is a template-rendering check, not a full Admin browser installation. Prepared with AI assistance and locally verified.

Frontend companion: https://github.com/getgrav/grav-plugin-form/pull/646


rhukster:
Thanks — worth noting Twig's `empty` test isn't PHP's `empty()`, so this keeps `0` while still dropping null/false/''. Merged.

PR #2522, opened 2026-09-13
fix: preview SVG files in file picker

### LINKED ISSUE #2510: Add svg preview to filepicker
Current state: Supported extensions include jpg, jpeg, png, gif, webp, avif. 
https://github.com/getgrav/grav-plugin-admin/blob/eb3cc2e9d1de698b11bcc2376e3c77ce31e07cad/themes/grav/js/admin.min.js#L4651
https://github.com/getgrav/grav-plugin-admin/blob/eb3cc2e9d1de698b11bcc2376e3c77ce31e07cad/themes/grav/app/forms/fields/filepicker.js#L83

Request: Please add support for SVG previews as well.

Current:

Image

Expected:

Image

sridharkalaibala:
SVG files currently appear without a preview in image-enabled file pickers because the renderer only recognizes raster image extensions. Add SVG to the same case-insensitive extension list so dropdown options and selected values use the existing `` preview path.

Fixes #2510.

Validation:
- Chromium 153 with the actual file-picker source and Selectize: 10 cases cover lowercase/uppercase SVG, PNG, non-image files, pending SVG, and previews disabled, for both option and selected-item rendering. The selected SVG loads with its expected intrinsic dimensions. Upstream fails the SVG cases; this change passes.
- `yarn@1.22.22 install --frozen-lockfile --ignore-scripts` and `npm run prod` pass. Build reports existing bundle-size warnings and an outdated Browserslist database notice.
- Updated the checked-in admin bundle. Removed unrelated generated comment-whitespace changes; final diff is one extension-list change in source and one in the bundle. Lockfile and vendor bundle are unchanged. `git diff --check` passes.

Browser validation uses a fixture for the file-list response and serves a small SVG; it is not a full authenticated Grav installation test. Existing preview settings, pending-file behavior and thumbnail URL selection are unchanged.


rhukster:
Thanks — the file picker was the odd one out here; `files.js` already listed svg, and uploads run through `sanitizeSVG()` before they land. Merged.

